PR #17, opened 2018-08-20
Adding rules page

ryelo:
This is a PR is to emulate the old Rules Page on Insights

Included is a new component: RulesCard as well as a Skeleton Loader when the component is loading asynchronously. When the component is initialized, the skeleton will load and wait until the API fills in the information. At that point, the skeleton disappears and the content loads.

When a Rule has a system hit, it links back to /actions/:type/:id. Which is the same secondary level actions page as the regular action flow.


ryelo:
@bmclaughlin I'm going to go ahead and merge this. Still review it and open up any issues you have with this when you get back.

PR #26, opened 2018-09-03
Allow lists to show with no items

dLobatog:
Currently, if I run this app just cloned from the repo, response.rules
is empty. Therefore when I visit the Actions or Rules pages I don't see
anything but an error message that you can't call .length on an
undefined method. This commit fixes that problem.


ryelo:
@dLobatog Good thinking. I'm okay with it as long as @bmclaughlin is. We should also have a card that says no info if no data is provided, but that can come in a later commit.

bmclaughlin:
@dLobatog, @ryelo,  looks good to me too.  Merging...

PR #29, opened 2018-09-06
Correctly names stylintrc.json, adds default warning

AllenBW:
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#loading-the-configuration-object

Valid name issssssss `.stylelintrc.json` 🤗


AllenBW:
@ryelo https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#defaultseverity

> The default severity level for all rules that do not have a severity specified in their secondary options.

ryelo:
Right, but why do we want any errors at all, even warnings? IMO, if you have an error of some kind, it should break the build and you should fix it instead of throwing up code with warnings.

Thoughts? @bmclaughlin

PR #33, opened 2018-09-13
Add filtering to Impacted Systems table on Rules page

bmclaughlin:
Added basic filtering to the Impacted Systems table on the Rules page.  Will duplicate effort on remaining pages once merged.


src/SmartComponents/Actions/ListActions.js
```diff
@@ -44,9 +46,12 @@ class ListActions extends Component {
         this.setPage = this.setPage.bind(this);
         this.setPerPage = this.setPerPage.bind(this);
         this.parseProductCode = this.parseProductCode.bind(this);
+        this.onSearch = this.onSearch.bind(this);
     }
 
     componentDidMount() {
+        document.getElementById('root').classList.add('actions__list');
```
AllenBW: ok I can stomach this... if we removed the class in the `componentDidUnmount` otherwise 😱thats gonna be one ova classy root after some time

AllenBW:
Mmmmmmmm two more things....

1.  at a modest resolution that search box is tiny tiny, might want it to be fixed width rather than responsive... 
screen shot 2018-09-17 at 3 34 37 pm

2. as no guidance on case sensitivity was provided, imma err on the side of prosumer, say we ought to have this be case insensitive search

bmclaughlin:
@AllenBW, done and done.

src/SmartComponents/Actions/_actions.scss
```diff
@@ -14,14 +14,16 @@
     }
 }
 
+.page__actions.actions__list {
+    table.impacted-systems-table { margin: $ins-margin 0; }
+    div.impacted-systems-filter-input { @include rem('width', 300px); }
```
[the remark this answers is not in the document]
bmclaughlin: That makes sense, thanks!

ryelo:
@bmclaughlin Can we use a combination of the stack layout for the wrappers and then grid for the filters instead of writing defined widths? This should eliminate some of the styles like:
```css
.page__actions.actions__list {
     table.impacted-systems-table { margin: $ins-margin 0; }
     div.impacted-systems-filter-input { @include rem('width', 300px); }
}
```

src/SmartComponents/Actions/_actions.scss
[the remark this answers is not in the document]
ryelo: see comment below about using the grid, because we'll probably add other filters, and this will future proof.

bmclaughlin:
@ryelo, @AllenBW I think this should be gtg now.

AllenBW:
jumping in!

src/SmartComponents/Actions/ListActions.js
```diff
@@ -39,10 +41,12 @@ class ListActions extends Component {
         this.setPage = this.setPage.bind(this);
         this.setPerPage = this.setPerPage.bind(this);
         this.parseProductCode = this.parseProductCode.bind(this);
+        this.onSearch = this.onSearch.bind(this);
```
AllenBW: outta curiosity, why are we doing this here (and everywhere else)?

AllenBW: ah got it, when a function relies on the component state (or external inputs), we gotta bind it

AllenBW: the binding is necessary to make `this` work in the callback

PR #35, opened 2018-09-17
Updating markup to reflect new linting rules

ryelo:
This changes markup, but is non breaking

## Curly Braces
These will apply to props and arrays inside of arrays.  Before, these only worked on the "top most level" (1) and wouldn't check if the curly braces were inside another array or inside of an object.

(1) Edit: Top level meaning everything outside of return statements

The only thing I'm still investigating is, for example, in RulesCard where we're setting HTML inside. I might just rewrite it and sanitize it, then use it as a variable. Ideally, if you need to use double curly braces, IMO it should be like

 `{{ foo }}`  instead of `{ { foo } }`

However, I don't hate the second way.

It might just need to a contribution to the React eslint plugin because I can't find anything about it anywhere which might end up with me opening a PR.

## Brackets
Brackets now follow similar rules to curly braces where they require a space inside.
Formerly: `[foo, bar]`
Proposed: `[ foo, bar ]`


.eslintrc.yml
```diff
@@ -83,6 +83,7 @@ rules:
     - error
     - always
     - objectsInObjects: false
+      arraysInObjects: false
```
AllenBW: should read `- arraysInObjects: false`?

.eslintrc.yml
```diff
@@ -71,6 +82,8 @@ rules:
   object-curly-spacing:
     - error
     - always
+    - objectsInObjects: false
+      arraysInObjects: false
```
AllenBW: should read `- arraysInObjects: false` ?

AllenBW:
> {{ foo }} instead of { { foo } }

I hate that second way 😏 🤗

.eslintrc.yml
ryelo: It should be like this because these exceptions are read as one attribute, even if there are two of them.

ryelo: Example: https://github.com/RedHatInsights/insights-advisor-frontend/blob/506bd384c4bc1902113de8b4445cf66620776b9f/.eslintrc.yml#L98

AllenBW:
Also, ps, who is deciding how or what we style? IMO we ought to just be done with it, (the act of styling and wiggling back n fourth on it) and just do what the pfr repo is doing. The configuration [here](https://github.com/patternfly/patternfly-react/blob/master/.prettierrc) and [here](https://github.com/patternfly/patternfly-react/blob/master/.stylelintrc) is **far** less than whats going on in this repo (yeah this might be something to bring up in the starter app also

ryelo:
@AllenBW Partially that is because they're being consumed by an NPM package and not being deployed as an app - especially the starter app. There are ~7-8 apps that, in the end, will be going to consuming the starter app, if we have consistent styling across all of them, it'll be easy for a person to be able to contribute anywhere and know the code standards.

AllenBW:
Cool so then why not get behind what pfr is doing rather considering we do/will consume a great deal of it?  It really boils down to avoiding prs like this (and all the future ones), [cuz these deps already have it figured out...](https://github.com/patternfly/patternfly-react/blob/master/package.json#L56-L60)

ryelo:
They're importing [airbnb's style](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base/rules) which is *much* more extensive than ours is.

Again, people consume pf as components through an npm package, not through a repo which they have to build on top of and other devs have to support.

ryelo:
@AllenBW we don't have to merge this. If we want to import another linter like airbnb's, that's fine. But as of right now, there are multiple places that break rules in this app. The only requirement is that if something breaks rule, it needs to throw an error and the app needs to break so that it doesn't hit the webhook that deploys it.

AllenBW:
This is the fastest way to whole @ryelo, am totally down to merge. Bringing up the discussion [over here](https://github.com/RedHatInsights/insights-frontend-starter-app/issues/46) cuz that's where it really belongs.

ryelo:
@AllenBW Mind rechecking the files you changes to make sure I hit everything and it's working correctly?

AllenBW:
jumping in, please stand by...

AllenBW:
The only thing is.... `Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration. `

`.eslintrc.yml`
```
settings:
  react:
    version: '16.0'
```

ryelo:
@AllenBW updated!

PR #39, opened 2018-09-19
Updates more deps

AllenBW:
fixes #38

I promise, totally done with deps (for a lil while) after this one 

### see look, doesn't that look identical and fantastic!
screen shot 2018-09-19 at 11 26 09 am

PR #42, opened 2018-09-20
Prevents undefined prop error  in listactions component

AllenBW:
I am guilty of not jumping into a pr, #39, and am now paying for the pony. Timey whimey component things are exciting especially when we're doing data processing  in the component, the more permanent fix is gonna be moving most of this logic into actions...

### whats it look like ?!!? but don't worry, 's fixed with this pr
screen shot 2018-09-20 at 5 13 43 pm

### DIFF
--- a/src/SmartComponents/Actions/ListActions.js
+++ b/src/SmartComponents/Actions/ListActions.js
@@ -31,7 +31,9 @@ class ListActions extends Component {
             cols: [ 'Type', 'Name', 'Reported' ],
             unfilteredRows: [],
             rows: [],
-            rule: {},
+            rule: {
+                description: ''
+            },
             sortBy: {},
             itemsPerPage: 10,
             page: 1


ryelo:
Instead of adding an `or ' '`  to the `PageHeaderTitle`, why not just declare the state inside the constructor like we do with other variables?

```
constructor(props) {
    super(props);
        this.state = {
            cols: [ 'Type', 'Name', 'Reported' ],
            unfilteredRows: [],
            rows: [],
            rule: {
                description: ''
            },
            sortBy: {},
            itemsPerPage: 10,
            page: 1
        };
    . . .
}
```

AllenBW:
Yeah that's another way to solve the problem, not sure its a better way 🤷‍♂️

ryelo:
@AllenBW 🤷‍♂️ I would say let's just keep it consistent for now since that's how we're setting the other state vars. @bmclaughlin ?

bmclaughlin:
@AllenBW @ryelo I vote to go with consistency till we find The Right :tm: solution.

PR #45, opened 2018-09-25
Adding widget attributes

ryelo:
QE wants us to add `widget-type='InsightsFoo'` to any presentational components on the top level tag. This way they can use their DOM scanner appropriately. 

If there are multiple instances of the same widget on a page, there needs to be a separate `widget-id` that can identify it.

These only need to be applies to Insights specific components. Patternfly is doing the same thing. I will also be updating the component library to follow the same pattern.

### DIFF
--- a/src/PresentationalComponents/RulesCard/RulesCard.js
+++ b/src/PresentationalComponents/RulesCard/RulesCard.js
@@ -48,7 +48,7 @@ const RulesCard =
     }
 
     return (
-        <Card { ...props } className = { rulesCardClasses }>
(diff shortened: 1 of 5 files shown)


AllenBW:
@ryelo is this a wip?

ryelo:
@AllenBW No, just got new information that I wanted to include before merge.

AllenBW:
Good to know we need to be adding `widget-type='InsightsFoo'` to some things, thanks for the heads up @ryelo !

PR #53, opened 2018-10-02
Fix aria-label on Impacted Systems

bmclaughlin:


src/PresentationalComponents/RulesCard/RulesCard.js
```diff
@@ -41,7 +41,7 @@ const RulesCard =
     if (hitCount > 0) {
         renderHitCount =
         <Link to={ `/actions/${category.toLowerCase()}/${ruleID}` }
-            aria-label='No impacted systems'> View Impacted Systems ({ hitCount }) </Link>;
+            aria-label={ `${hitCount} impacted system${hitCount > 1 ? 's' : ''}` }> View Impacted Systems ({ hitCount }) </Link>;
```
AllenBW: so this would read 
`12 impacted systems` 
or
`0 impacted system`

☝️ kinda reads odd, Imma thinking it should always be plural .

ryelo: if it was 0, it wouldn't render that, it'll render line 47 that says `No impacted Systems`. The singular should only happen if it is `1`.

AllenBW: so then it should read { `${hitCount} impacted system${hitCount === 1 ? '' : 's'}` }

bmclaughlin: @AllenBW Six or a half dozen?  I'm testing for the most common case of the conditional and using less characters.

PR #56, opened 2018-10-03
Add breadcrumbs

bmclaughlin:


AllenBW:
@bmclaughlin can ya upload some ss?  though they don't go nogo on prs yet, I could imagine this one being something ux might want to 👀

AllenBW:
Looks to be missing on the listrules component 🤔

src/PresentationalComponents/Crumbs/Crumbs.js
```diff
@@ -0,0 +1,70 @@
...
+
+    render() {
+        const items = this.buildItems();
+        return (
+            <React.Fragment>
+                {
+                    items.length > 0 && <ol className='ins-breadcrumbs' widget-type='InsightsBreadcrumbs'>
+                        { items.map((link) => (
+                            <li key={ link.navigate }>
+                                <Link to={ link.navigate }>{ link.title }</Link>
```
AllenBW: probably want to give this a `aria-label`

AllenBW:
Taking two steps back, is there a reason why we aren't using [pfr breadcrumbs here?](https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=patternfly-react%2FNavigation%2FBreadcrumb&selectedStory=Breadcrumb&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs)

AllenBW:
ok ok ok i can do one better its not the pfr one... but at the very least we ought to be consuming what the rest of insights is gonna do... https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/breadcrumbs.md

ryelo:
We should be consuming the breadcrumbs from the front end components

bmclaughlin:
Was trying to find a good way to tie the breadcrumbs into react-router without much luck. Currently the markup (except the missing aria-label, doh!) is the same as the Breadcrumbs component with the exception of replacing the anchor tags with Link tags from react-router.  I'll take another look.

AllenBW:
gonna add aria-label to front end things

karelhala:
@bmclaughlin if you find good way to combine breadcrumbs and `react-router` together that would be awesome! I tried that and failed mid way. But anyways, you should use `breadcrumbs` component and call `history.push()` when `onNavigate` action is triggered so you don't have to write the markup and everything once again.

bmclaughlin:
@karelhala, Yeah, I probably got to, or close to, the same point you did.  I'm currently updating the PR as you mentioned.

bmclaughlin:
@AllenBW @ryelo, when you have a moment could you take a look?

ryelo:
Presentation looks good!

I think just looking over very briefly, if you could contribute those styles back up to Insights Frontend Components, that would be awesome 😄

AllenBW:
This should be present on all views right? Rules list included?

bmclaughlin:
@AllenBW The conundrum I'm running into is that Rules is only one page, and as such doesn't have breadcrumbs like the Actions Overview.  If we roll back the root of the breadcrumbs to make Advisor the root of the URLs it would always link to Actions Overview which feels wrong when on the Rules page.

And just throwing the Current page up there feels weird, but I think that'd be better than rolling back the root to Advisor.  Thoughts?

ryelo:
From a UX perspective, the current design should not use breadcrumbs, because there is no way to get to the rules page from the actions overview. If you nested the Rules page underneath actions overview (which doesn't really make sense unless you could hit a rules widget), then that would warrant breadcrumbs. 

Because the breadcrumbs represent a sitepath, the rules page should not have breadcrumbs in the current state. However, if you were to hit actions 2 from the rules page, then actions 2 would have breadcrumbs that would represent that such as `Rules > foo `

AllenBW:
So Advisor has two left nav entries `Actions` and `Rules`  (as base defaults to actions, it doesn't count) bothof these locations act as roots from which other views are branched... every root should have a breadcrumb, a starting point, (unless that's not true)... 

I'd feel strongly enough about this to bring in UX arbiter.. my bottom line, surprise breadcrumb building should never occur, its gradual, each navigation location should contribute a piece.

ryelo:
Having breadcrumbs on top level pages is redundant because if you're on a top level page (overview or rules), the breadcrumbs aren't clickable, which defeats the purpose of breadcrumbs. 

Actions Overview and Rules are both items in the navigation. Thus, no breadcrumbs should be on rules due to the fact you don't actually have to hit actions to get to rules. 

If you're on another application, say the Dashboard, and you don't really care about hitting the advisor overview, and all you care about is the rules page, if you click rules in the navigation, then there's no point of saying `actions > rules` because you never hit actions in the first place.

ryelo:
Speaking on the point above, when the new navigation specs are done, I don't believe that the "advisor" tab in the navigation will direct you to the top level of the page. It will be a dropdown which will make you select rules or overview.

AllenBW:
Ok fair enough, I found a new issue to whine about, attached gif shows the navigation working correctly, but the textual representation to be a bit off
test

ryelo:
@AllenBW That gif above is incorrect. The breadcrumbs should flow from rules in that situation and properly say `Rules > foo`

bmclaughlin:
@karelhala I could replace URL references with the `match` object, I'm still going to run into not being able to know if I came from `Rules` or `Actions` (no history object lets  you inspect or iterate over history that I've found).  I'm now looking into storing breadcrumb data in `AdvisorStore` so that we have nuanced control over naming and pathing.

src/SmartComponents/Actions/ListActions.js
```diff
@@ -154,11 +157,40 @@ class ListActions extends Component {
         });
     }
 
+    onNavigate(_event, _item, key) {
+        const { history } = this.props;
+        history.go(-key);
+    }
+
+    parseBreadcrumbs(breadcrumbs, params) {
```
AllenBW: If there is a possibility this might be used again, parsing the breadcrumb title, we ought to abstract the logic into a helper or a functional component...

src/SmartComponents/Actions/ViewActions.js
```diff
@@ -138,11 +140,21 @@ class ViewActions extends Component {
         return parsedTitle.length > 1 ? `${parsedTitle[0]} ${parsedTitle[1]} Actions` : `${parsedTitle}`;
     }
 
+    onNavigate(_event, _item, key) {
```
AllenBW: also this probably should be abstracted into the same helper, imported into the file that uses it, as code climate would say... duplicate code found 😩

bmclaughlin:
@AllenBW @ryelo @karelhala When you get a chance, please take a look.

src/SmartComponents/Actions/ViewActions.js
```diff
@@ -188,13 +195,20 @@ class ViewActions extends Component {
 }
 
 ViewActions.propTypes = {
-    match: PropTypes.any,
+    breadcrumbs: PropTypes.array,
     fetchRules: PropTypes.func,
+    history: PropTypes.object,
```
AllenBW: last thing then i promise we can merge this, cleaning up the props here... we dont need history path params

src/SmartComponents/Actions/ListActions.js
```diff
@@ -217,11 +229,18 @@ class ListActions extends Component {
 }
 
 ListActions.propTypes = {
+    breadcrumbs: PropTypes.array,
+    history: PropTypes.object,
```
AllenBW: last thing then i promise we can merge this, cleaning up the props here... we dont need history path params

src/SmartComponents/Actions/ListActions.js
```diff
@@ -217,11 +229,18 @@ class ListActions extends Component {
...
     match: PropTypes.any,
+    path: PropTypes.string,
+    params: PropTypes.object,
     AdvisorStore: PropTypes.object
 };
 
 const mapStateToProps = (state, ownProps) => ({
+    breadcrumbs: state.AdvisorStore.breadcrumbs,
+    path: state.routerData.path,
```
AllenBW: don't need path or params here either

src/SmartComponents/Actions/ViewActions.js
```diff
@@ -188,13 +195,20 @@ class ViewActions extends Component {
...
+    path: PropTypes.string,
+    params: PropTypes.object,
     rulesFetchStatus: PropTypes.string,
     rules: PropTypes.object
 };
 
 const mapStateToProps = (state, ownProps) => ({
+    breadcrumbs: state.AdvisorStore.breadcrumbs,
+    path: state.routerData.path,
+    params: state.routerData.params,
```
AllenBW: don't need path or params here either

PR #58, opened 2018-10-05
Use router params and set redux registry from frontend component

karelhala:
This PR is blocked by https://github.com/RedHatInsights/insights-frontend-components/pull/116

* Use router params - QE has easier definition where they are located so they have much easier job.
* Use component's registry - this way we don't have to take care of our registry, but this can be maintained from component's registry. So any changes did to this registry will take effect in this one as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1505,7 +1505,7 @@
     },
     "ansi-colors": {
       "version": "1.1.0",
-      "resolved": "http://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
+      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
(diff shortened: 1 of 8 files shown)


src/Store/index.js
```diff
@@ -1,52 +1,13 @@
-import { applyMiddleware, combineReducers, createStore, compose } from 'redux';
 import promiseMiddleware from 'redux-promise-middleware';
 import { AdvisorStore } from '../AppReducer';
-
+import { getRegistry } from '@red-hat-insights/insights-frontend-components';
 let registry;
 
-class ReducerRegistry {
```
AllenBW: Hard nope on removing this. Redux dev tools does NOT work when pulling in reducer registry.  Till it does, we gotta run with our own.

AllenBW: Oh awesome! All the same, I would much prefer we unwrap that additional change from this pr, do it standalone if indeed it does fix the issue.... 🙇‍♀️

karelhala: Sure thing, good thinking.

AllenBW:
Looks like the other one was merged, jumping into this right now!

PR #60, opened 2018-10-08
Listrules support api pagination

AllenBW:
Does a little bitta work on RHIADVISOR-209

You'll notice there setting the page can be a little difficult, this isn't an issue with this work, just the way the component is built... it ought to have a bitta `debounce`... something to look into in the future...

### for the ~~lazy~~ mildly curious
test


AllenBW:
@bmclaughlin Yep the component could definitely use some polish, but thats not the work done in this pr, nor should it be, alls this does is gets the pagination workin with the api.  The debounce absence bug, edge case pagination button disable bug, all work to be done in other prs, doesn't fall within the scope of this one 🤗

PR #61, opened 2018-10-09
Adds missing attributes impact, likelihood

AllenBW:
### Completed
* https://github.com/RedHatInsights/insights-advisor-api/pull/87
* https://github.com/RedHatInsights/insights-advisor-api/pull/98


AllenBW:
The last two attribute we need before this can be closed out is 
* resolution_risk
* risk_of_change

AllenBW:
re-wipping till other to attributes are added

AllenBW:
ok it looks like `resolution_risk` should actually be `severity` (looking at classic)

AllenBW:
Not an issue, latest ui components fixes this, unsupported battery values,
will make a deep update pr a little later today

PR #63, opened 2018-10-11
Upgrade deps, specifically babel 6 to 7

AllenBW:
Includes a bunch of cleanup, all npm scripts nominal 

There's a bunch of neat lil changes here, so if you have questions please do ask.  <3


AllenBW:
And it goes without saying, if we're ok with this, I'll update the the insights-frontend-starter-app

PR #66, opened 2018-10-18
Locking down deps to ensure predictable app behavior

AllenBW:
Last thing we need on a sass product is dep surprises crashing the app.  Locking it  down to save our future sanity.


kensible:
I haven't pulled it down to test but I approve of the approach and agree with the reasoning behind it.

PR #77, opened 2018-10-26
[WIP] Add Filters component

bmclaughlin:
Early stages, comments welcome!


src/SmartComponents/Filters/RiskOfChange.js
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
```
AllenBW: Risk of change and total risk look eerily similar is there a generic dropdown component that could be passed an options to represent both? Does such a component exist in either pfr or frontend-component libraries?

bmclaughlin: Plenty of refactoring rounds still to come, that's on the top of my mind too.  I figure there will be an options hash passed to differentiate those filters.

PR #78, opened 2018-10-29
Replaces depreciated webpack-serve and reconfigures confg, updates deps

AllenBW:
This one is kinda neat though, cuz this brings **ALL DEPS** up to date, that last one, reworking webpack-serve to use webpack-dev-serve is a little tricky, and there's still a few bugs to work out, but here's where we're going with it!

### proof its workin 😏 
screen shot 2018-10-29 at 4 04 55 pm


config/dev.webpack.config.js
```diff
@@ -3,17 +3,13 @@
...
-    content: config.paths.public,
+webpackConfig.devServer = {
+    contentBase: config.paths.public,
+    hot: true,
     port: 8002,
-    dev: {
-        publicPath: '/insights/platform/advisor'
-    },
-    // https://github.com/webpack-contrib/webpack-serve/blob/master/docs/addons/history-fallback.config.js
-    add: app => app.use(convert(history({})))
```
AllenBW: Best I can tell, this was for preserving browser history? or something?  Anyhow appears to work just fine when using `historyApiFallback: true` @ryelo @karelhala either of yah intimately familiar with this?

karelhala: @AllenBW yeah, you are right. This was mainly to preserve browser history and enable refreshing when new changes to code happend.

AllenBW:
@karelhala awesome thanks for the 👀 will add to the starter app some point today!

PR #84, opened 2018-11-01
Ends up we need to display reports not rules

AllenBW:
Thanks @dkuc!

### looks like....
screen shot 2018-11-01 at 2 37 04 pm


AllenBW:
@dkuc will push to prod as soon as we merge this 👍

PR #90, opened 2018-11-06
Declutters unused configs

AllenBW:
More for a topic of discussion... thoughts on cleaning this up a bit?  Can always add em back if/when additional plugins are necessary


bmclaughlin:
Looks good to me, everything runs normally with the changes.  We can always re-implement if/when needed.  @ryelo, any thoughts?

PR #95, opened 2018-11-15
Ensures our bar chart filtering of the rule list behaves expectedly

AllenBW:
If the data in the list doesn't show up, it might be that the api hasn't updated yet, but check the network tab and you'll notice it lines up just okie dokie with https://github.com/RedHatInsights/insights-advisor-api/pull/120#issuecomment-438731447

### DIFF
--- a/src/SmartComponents/Actions/ViewActions.js
+++ b/src/SmartComponents/Actions/ViewActions.js
@@ -1,3 +1,4 @@
+/* eslint camelcase: 0 */
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import {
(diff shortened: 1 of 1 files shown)


AllenBW:
Thanks for the review! Made some changes as per your comments... 🤗 🏹

src/SmartComponents/Actions/ViewActions.js
```diff
@@ -50,14 +51,30 @@ class ViewActions extends Component {
 
     componentDidMount () {
         document.getElementById('root').classList.add('actions__view');
-        this.props.fetchRules({ page_size: this.state.itemsPerPage, category: this.props.match.params.type }); // eslint-disable-line camelcase
+        const riskMap = {
```
AllenBW: I could see exporting this to the constants file possibly.... if it gets used again somewhere...

bmclaughlin: I don't see another match for it at the moment, maybe in the future?

PR #96, opened 2018-11-16
Reconfigure status messages to use api response

AllenBW:
Past: canned response for all error messages
Present: lets display the (sometimes ugly) api response

(i know i know, why not just render the html in the error body, nope not an option.  technically doable but defeats the purpose of the pf design)

### sometimes ugly but still helpful
screen shot 2018-11-16 at 11 33 40 am 

### sometimes helpful/pretty
screen shot 2018-11-16 at 11 30 31 am


bmclaughlin:
A future enhancement might be stripping the HTML tags out of the error messages?

PR #104, opened 2018-12-07
add Remediation support (via hot-loaded RemediationWizard)

jharting:
Note that at this point Remediations only has API deployed in CI so CI is the only environment where a remediation creation will actually work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1315,9 +1315,9 @@
       },
       "dependencies": {
         "core-js": {
-          "version": "2.5.7",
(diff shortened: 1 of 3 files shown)


jharting:
Also, some styling work will be needed to place the button to the right spot. Feel free to place it wherever makes sense.

AllenBW:
for the curious, here's a ss of the action, 

screen shot 2018-12-10 at 10 05 32 am

@jharting gonna throw button next to the ansible icon, commit incoming

jharting:
@AllenBW sounds good!

PR #107, opened 2018-12-11
Refactors dashboard doughnut to use pf4 widget

AllenBW:
Donut? Doughnut? 

screen shot 2018-12-12 at 1 20 35 pm

### The bummers
* mousing over the legend (or chart) does not slightly enlarge the doughnut (totally possible I believe... just passing an `onMouseover` event  to the ChartLegend.events.... something to look into)
* label is not in donut center, this is tricky, [pf4 pie chart fixes the size of the chart and label and overlays the latter](https://patternfly-react.surge.sh/patternfly-4/components/donutchart#ChartLabel)

can work on all the bummers, but wanted to get this thrown out there first, cuz peeps be complain' bout da lookz of present state 💃 🦂

### DIFF
--- a/src/PresentationalComponents/Charts/ActionsOverviewDonut.js
+++ b/src/PresentationalComponents/Charts/ActionsOverviewDonut.js
@@ -1,0 +1,82 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+import { ChartDonut, ChartLabel, ChartLegend, ChartTheme } from '@patternfly/react-charts';
+import { Grid, GridItem } from '@patternfly/react-core';
+
+class AdvisorOverviewDonut extends React.Component {
+    getChart = (theme, typeNames) => (
+        <ChartDonut
+            data={ this.props.category.map((value, key) => ({ x: typeNames[key], y: value, label: `${typeNames[key]}: ${value}` })) }
+            theme={ theme }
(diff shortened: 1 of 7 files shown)


src/PresentationalComponents/Charts/ActionsOverviewDonut.js
```diff
@@ -0,0 +1,82 @@
...
+                    verticalAnchor="middle"
+                />
+
+            </div>
+        );
+        const typeNames = [ 'Availability', 'Security', 'Stability', 'Performance' ];
+
+        return (
+            <Grid gutter="lg">
+                <GridItem lg={ 1 }/>
```
ryelo: What's the purpose of using these empty grid items?

AllenBW: 🤷‍♂️ make it kinda be in the center kinda, else it gets shoved over to the right (i'd not believe it unless i didn't see it 😭 )

ryelo: @AllenBW could you use the bullseye or just` margin-left/right: auto`?

AllenBW: well it was a good thought.... bullseye does this
screen shot 2018-12-13 at 6 44 56 am

as for adding css 😬

PR #111, opened 2018-12-17
Update hotloading inventory as per new docs

AllenBW:
updated inventory component means details view is no longer a tab, its it's own page https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/inventory.md#usage-of-hot-loading

[**to get this alive follow the steps drawn out here** 
](https://github.com/RedHatInsights/insights-advisor-frontend/pull/100)
### whats this look like? 
screen shot 2018-12-17 at 12 42 44 pm
screen shot 2018-12-17 at 12 42 38 pm
screen shot 2018-12-17 at 12 45 26 pm

### DIFF
--- a/src/PresentationalComponents/Inventory/InventoryDetails.js
+++ b/src/PresentationalComponents/Inventory/InventoryDetails.js
@@ -1,0 +1,47 @@
+import React from 'react';
+import * as reactRouterDom from 'react-router-dom';
+import * as reactCore from '@patternfly/react-core';
+import * as reactIcons from '@patternfly/react-icons';
+import { Main, registry as registryDecorator } from '@red-hat-insights/insights-frontend-components';
+import Loading from '../../PresentationalComponents/Loading/Loading';
+import { entitiesDetailsReducer } from '../../AppReducer';
+
+@registryDecorator()
+class InventoryDetails extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            InventoryDetails: () => <Loading />
+        };
+
+        this.fetchInventoryDetails();
+    }
+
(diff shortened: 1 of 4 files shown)


src/PresentationalComponents/Inventory/InventoryDetails.js
```diff
@@ -0,0 +1,49 @@
...
+
+        this.setState({
+            InventoryDetails: inventoryConnector().InventoryDetail
+        });
+    }
+
+    render() {
+        const { InventoryDetails } = this.state;
+        return (
+            <React.Fragment>
```
ryelo: This shouldn't need a fragment

ryelo: Nitpick 😄

bmclaughlin: Good catch!

AllenBW: yup good point, yet another leftover from building

src/PresentationalComponents/Inventory/InventoryDetails.js
```diff
@@ -0,0 +1,47 @@
...
+        this.setState({
+            InventoryDetails: inventoryConnector().InventoryDetail
+        });
+    }
+
+    render() {
+        const { InventoryDetails } = this.state;
+        return (
+            <Main>
+                <InventoryDetails />
```
karelhala: You are missing root in here so the back button works correctly. Sorry probly missed that in docs. The root for you should be `/actions/:type/:id` since that is from where you are navigating to inv detail.

PR #118, opened 2018-12-20
Ensure correct kba link is used for rules

AllenBW:
We no go off and fetch the kba view_uri using the node_id to ensure we're getting the right one every time, won't show the link if there isn't anything to show

Followup work will transfer this same thing to the inventory rule list

### Whats it look like? see that link, bottom left? its CORECT 🤤 
screen shot 2018-12-20 at 1 59 56 pm


AllenBW:
thanks @dkuc for your help with this 🤗 🙇

PR #119, opened 2018-12-20
Adds pagination/filtering to inventory table

AllenBW:
[waiting on this to show up in chrome before we merge](https://github.com/RedHatInsights/insights-frontend-components/pull/168)


karelhala:
WIP can be removed, this update will not break current inventory.

PR #121, opened 2018-12-20
Filter viewactions on severity rather than res_risk

AllenBW:


AllenBW:
@dkuc given that https://github.com/RedHatInsights/insights-advisor-api/pull/158 was merged, we ok to push this one in?

PR #148, opened 2019-01-18
Adds breadcrumbs and additional visual enhancements to inv details

AllenBW:
### here's what we look like!! breadcrumbs by hostname when available, uuid when not 
screen shot 2019-01-22 at 3 25 59 pm
screen shot 2019-01-22 at 3 25 45 pm


karelhala:
FYI once chrome is updated (should be at the moment) your tabs will change from those underline texts to boxed tabs.

PR #150, opened 2019-01-22
Revert "Refactor breadcrumbs to make them match based"

AllenBW:
Reverts RedHatInsights/insights-advisor-frontend#149

This pr ended up being a regression, upon refreshing the page, breadcrumbs will break the app, `Uncaught TypeError: Cannot read property 'navigate' of undefined`


AllenBW:
cc @bmclaughlin

PR #157, opened 2019-01-24
fix remediations integration

jharting:


jharting:
@AllenBW looks like some regressions happened i.e. removal of these function bindings https://github.com/RedHatInsights/insights-advisor-frontend/commit/477a5db8c7e7d84c08ea7a183097edb8c65ec3b7#diff-fd62c8564a524f7b19ab878b81bc8f9dL28 makes Remediations integration no longer work.

PR #160, opened 2019-01-25
BU directed the removal of vulnerabilities tab

AllenBW:
### see its gone now!
screen shot 2019-01-25 at 10 25 33 am


AllenBW:
@ncameronbritt thats how the inventory details component work, we can't not have the tab at this time cc @karelhala

AllenBW:
@karelhala maybe we could do something along the lines of... if the component doesn't have a title, and is the only active app, there is no tab for it?

AllenBW:
@ncameronbritt so after that last dep update, this is what we have, no tab!
screen shot 2019-01-27 at 9 31 33 am

ncameronbritt:
Thanks for the update @AllenBW!

PR #168, opened 2019-01-29
Forces fixed victory container height for donut so we don't 😱🍩

AllenBW:
Fixes https://projects.engineering.redhat.com/browse/RHIADVISOR-378
Fixes https://projects.engineering.redhat.com/browse/RHIADVISOR-371

### It looks like this
donut


bmclaughlin:
:+1:

PR #172, opened 2019-02-04
Update page layout to support chrome pf v1.0.175

AllenBW:
In support of https://github.com/RedHatInsights/insights-chrome/pull/147


src/Routes.js
```diff
@@ -28,7 +28,7 @@ const paths = {
 const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
     const root = document.getElementById('root');
     root.removeAttribute('class');
-    root.classList.add(`page__${rootClass}`, 'pf-l-page__main');
```
karelhala: No need to remove this line, we need to keep both classes there untill chrome PR gets merged.

PR #174, opened 2019-02-05
Update deps

AllenBW:


package.json
```diff
@@ -10,10 +10,10 @@
   },
   "dependencies": {
     "@babel/runtime": "7.3.1",
-    "@patternfly/patternfly-next": "1.0.171",
-    "@patternfly/react-core": "1.51.4",
-    "@patternfly/react-icons": "2.10.3",
-    "@red-hat-insights/insights-frontend-components": "3.33.48",
+    "@patternfly/patternfly-next": "1.0.175",
```
ryelo: Just FYI, we just found out that this will be moving to `@patternfly/patternfly`

bmclaughlin: :+1:

PR #180, opened 2019-02-05
Adds search and filter capability to rulelist/viewactions components

AllenBW:
spiritual successor to closes #173 closes #165 

#### Biggies
- ~not applied to actions page, won't be too tough~
- ~removing filters, we're waiting for https://projects.engineering.redhat.com/browse/RHIADVISOR-411~
- ~selecting multiples under the same title, right now filtering isn't additive, we're waiting for https://projects.engineering.redhat.com/browse/RHIADVISOR-411~
- ~closing and reopening filter dropdown does not preserved selected values~
#### Minor ones
- styling on filter button is off, but this worked earlier, so might be a chrome thing? unsure, for sure all the correct classes are being applied (i think we can hold off on this one for now, followup pr to fix once this is out there, can understand the issue more)

### What works
- filtering rules by all values dropdown values(across categories, the last value of each selected)
- all other input is additive/subtractive, impacted and  searching are done on a subset of the dropdown filters
- paging and limit of all search results
- uses all pf4 components, follows pf4 toolbar sample
- applied to actions page
- removing filters
- selecting multiple filters
- closing and reopening filter dropdown does preserve selected values

### and its supah sweet
filter

### DIFF
--- a/src/PresentationalComponents/Filters/CheckboxFilter.js
+++ b/src/PresentationalComponents/Filters/CheckboxFilter.js
@@ -1,0 +1,52 @@
+import React, { Component } from 'react';
+import { Checkbox } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+
+class CheckboxFilter extends Component {
+    state = {
+        checked: false
+    };
+
+    componentDidMount () {
+        this.props.filters[this.props.param] && this.props.filters[this.props.param].includes(this.props.value) ?
+            this.setState({ checked: true }) :
+            this.setState({ checked: false });
+    }
+
+    handleChange = checked => {
+        this.setState({ checked: !this.state.checked });
+        this.props.addRemoveFilters(
+            event.target.value,
+            event.target.getAttribute('param'),
+            checked
+        );
+    };
+
+    render () {
+        return (
+            <Checkbox
+                aria-label={ this.props.label }
+                id={ this.props.id }
+                isChecked={ this.state.checked }
+                label={ this.props.label }
+                onChange={ this.handleChange }
+                param={ this.props.param }
+                value={ this.props.value }
+            />
+        );
+    }
+}
+
+CheckboxFilter.propTypes = {
+    addRemoveFilters: PropTypes.func,
+    className: PropTypes.string,
+    currentPage: PropTypes.string,
+    id: PropTypes.string,
+    label: PropTypes.string,
+    param: PropTypes.string,
+    value: PropTypes.string,
+    filters: PropTypes.object
+};
+
+export default CheckboxFilter;
+
--- a/src/PresentationalComponents/Filters/CheckboxFilter.js
+++ b/src/PresentationalComponents/Filters/CheckboxFilter.js
@@ -16,8 +16,8 @@ class CheckboxFilter extends Component {
     handleChange = checked => {
         this.setState({ checked: !this.state.checked });
         this.props.addRemoveFilters(
-            event.target.value,
-            event.target.getAttribute('param'),
+            this.props.value,
+            this.props.param,
             checked
         );
     };
--- a/src/PresentationalComponents/Filters/DropdownFilters.js
+++ b/src/PresentationalComponents/Filters/DropdownFilters.js
@@ -1,0 +1,104 @@
+import React, { Component } from 'react';
+import { Dropdown, DropdownToggle } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+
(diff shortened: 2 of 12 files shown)


AllenBW:
> I think the result count should be the same line, it's not shown on a separate line in any of the mockups that I've seen.
good question, so this, checkout https://patternfly-react.surge.sh/patternfly-4/demos/toolbar results on a separate line is a pf4 design thing, mockups are kinda in the wrong here, Kendra signed off on this

karelhala:
I was trying to do something similiar with filters in inventory, but that got removed because of not that great UX design. We should contact PF folks and ask them to design dropdown menu with multiple context menus. Something like this:
dropdown-context

And potentially use https://pf4.patternfly.org/components/OptionsMenu/examples/ checkboxes. But this is blocked on PF core side because the'd have to split dropdown into 2 components and rewrite bunch of code. So this design is sufficient for now, just keep in mind that on smaller screens that will not be user friendly. Also once this gets merged can you @AllenBW move it to PF components so others can benefit from this?

KendraMar:
@AllenBW Yeah, we are sorry that the mocks are wrong. We did our mocks before the Pfly4 team did their table component design (we finished in Dec, they finished late Jan). Def go with whatever the Pfly4 table component tells you.

Going forward, UX will be using a mockup tool (Sketch) with (many) components created directly by the PatternFly4 design team. Hopefully that'll eliminate red herrings.

AllenBW:
for posterity, followup to this will include https://ansible.slack.com/archives/CADT032B0/p1550090092189700

src/PresentationalComponents/Filters/CheckboxFilter.js
```diff
@@ -0,0 +1,52 @@
...
+    componentDidMount () {
+        this.props.filters[this.props.param] && this.props.filters[this.props.param].includes(this.props.value) ?
+            this.setState({ checked: true }) :
+            this.setState({ checked: false });
+    }
+
+    handleChange = checked => {
+        this.setState({ checked: !this.state.checked });
+        this.props.addRemoveFilters(
+            event.target.value,
```
karelhala: Event is missing in attributes. But it's not really needed you can access param and value from props.

src/PresentationalComponents/Filters/DropdownFilters.js
```diff
@@ -0,0 +1,104 @@
...
+    onToggle = isOpen => {
+        this.setState({
+            isOpen
+        });
+    };
+
+    render () {
+        const { hideCategories, filters } = this.props;
+        const { isOpen } = this.state;
+        const filterData = [
```
karelhala: Can you move this const to special file or at least above component definiton? This way we will not recreate these values everytime we re re-render component.

src/PresentationalComponents/Filters/Filters.js
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    removeFilter = (key, value) => {
+        const newFilter = { [key]: this.props.filters[key].split(',').filter(item => item !== value).join(',') };
+
+        if (newFilter[key].length) {
+            this.props.setFilters({ ...this.props.filters, ...newFilter });
+            this.props.fetchAction({ ...this.props.filters, ...newFilter });
+        } else {
+            const filter = Object.assign({}, this.props.filters);
```
karelhala: No need to use `Object.assign`: `const filter = Object.assign({}, this.props.filters);` -> `const filter = { ...this.props.filters };`

src/PresentationalComponents/Filters/Filters.js
```diff
@@ -0,0 +1,104 @@
...
+            delete filter[key];
+            this.props.setFilters(filter);
+            this.props.fetchAction(filter);
+        }
+    };
+
+    render () {
+        const { children, searchPlaceholder, filters, resultsCount, hideCategories } = this.props;
+        return (
+            <Toolbar className='pf-u-justify-content-space-between pf-u-ml-xl pf-u-my-md'>
```
karelhala: Maybe consider using [insights-frontend-components/tableToolbar.md](https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/tableToolbar.md)

AllenBW: YEAH! good call, will be following up with another pr to do that.. right now its not a copy and paste fix, @ryelo mentioned it in slack, a the comment https://github.com/RedHatInsights/insights-advisor-frontend/pull/180#issuecomment-463364101 calls it out

PR #183, opened 2019-02-08
Fixes breadcrumb using rule.id

bmclaughlin:
Refactored Breadcrumb is now class-based and reduces the dependencies required of parent components.  On Inventory page, breadcrumbs now use the Rule description instead of the id 

Fixes RHIADVISOR-401

Before:
breadcrumbs-before

After:
breadcrumbs-after

### DIFF
--- a/src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
+++ b/src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
@@ -47,7 +47,7 @@ class Breadcrumbs extends React.Component {
         }
 
         // add :type breadcrumb (exception: Rules based breadcrumbs)
-        if (match.params.id !== undefined && crumbs[0].title !== 'Rules') {
+        if (match.params.type !== undefined && match.params.id !== undefined && crumbs[0].title !== 'Rules') {
             const title = this.getReadableType(match.params.type);
             crumbs.push({
                 title,
@@ -55,7 +55,7 @@ class Breadcrumbs extends React.Component {
             });
         }
 
-        if (match.params.inventoryId !== undefined) {
+        if (match.params.id !== undefined && match.params.inventoryId !== undefined) {
             crumbs.push({
                 title: this.props.rule.description,
                 navigate: crumbs[1].navigate + '/' + match.params.id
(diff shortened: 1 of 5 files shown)


src/SmartComponents/Actions/ListActions.js
```diff
@@ -4,7 +4,7 @@ import { Ansible, Battery, Main, PageHeader, PageHeaderTitle, RemediationButton,
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { Card, CardBody, CardHeader, Grid, GridItem, Stack, StackItem } from '@patternfly/react-core';
-import Breadcrumbs, { buildBreadcrumbs } from '../../PresentationalComponents/Breadcrumbs/Breadcrumbs';
+import Breadcrumbs from '../../PresentationalComponents/Breadcrumbs/Breadcrumbs';
```
ryelo: Is there any way we could use the breadcrumbs from patternfly instead?

ryelo: Nevermind. I see it.

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
```diff
@@ -1,52 +1,113 @@
...
+        return (
+            <React.Fragment>
+                { (ruleFetchStatus === 'fulfilled' || items.length > 0) && (
+                    <Breadcrumb>
+                        { items.map((oneLink, key) => (
+                            <BreadcrumbItem key={ key }>
+                                <Link to={ oneLink.navigate }>{ oneLink.title }</Link>
+                            </BreadcrumbItem>
+                        )) }
+                        <BreadcrumbItem to='' isActive>{ this.props.current }</BreadcrumbItem>
```
ryelo: nitpick: You don't need a `to=''` here

src/SmartComponents/Actions/ListActions.js
bmclaughlin: In the local Breadcrumbs component we're using PF4 components.  The local component just houses our logic.

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
bmclaughlin: Good catch, updated!

bmclaughlin:
@AllenBW, turns out that was the page and breadcrumb title.  Wrapped the output to display a loading tile to resolve.  The tile was less visually jarring imo, let me know what you think.

PR #205, opened 2019-02-21
Add Sort by Total Risk to Rules page

bmclaughlin:
Closes RHIADVISOR-311.  Should work soon since https://github.com/RedHatInsights/insights-advisor-api/pull/189 has been merged.

sort-by-total-risk


AllenBW:
@ncameronbritt here's what it looks like expanded... it does not appear that patternfly4 has a non table descriptive way to sort items... was hoping you could weigh in here on the design

screen shot 2019-02-21 at 11 29 54 am

AllenBW:
A few other thoughts.  Is there value to a non-sorted state? Maybe by default we should sort descending.  🤔 If the requirement to sort by other values should arise, how would that be handled?

src/SmartComponents/Rules/_rules.scss
```diff
@@ -1,3 +1,3 @@
 .page__rules {
-    // Filler
+    .rulesHideHitsGroup { margin-left: auto; }
```
AllenBW: I think we should be sticking to pf styles as much as possible, `pf-u-justify-content-space-between` removed on line https://github.com/RedHatInsights/insights-advisor-frontend/pull/205/files#diff-e0d9c4570235d8fddea19d06db87d149L48 achieved this

ncameronbritt:
@AllenBW Check out the sort widget here:
[](https://pf4.patternfly.org/components/Toolbar/examples-full/?component=Toolbar%20w/%20Pagination%20and%20expanded%20menus%20for%20bulk%20selector,%20sort,%20action%20menu,%20and%20pagination)

Eventually this page will use the table component, so we'll have those headers... :-)

cc @katierik

src/SmartComponents/Rules/_rules.scss
bmclaughlin: I took this approach looking at the pf4 simple toolbar sample markup [1].  The original `pf-u-justify-content-space-between` worked perfectly when there were two distinct sets to work with (inside of `Filters` and 1 child passed in), but it's not as clean when passing in multiple children and treating them differently).

[1] https://pf4.patternfly.org/components/Toolbar/examples-full/?component=Toolbar%20simple

AllenBW:
@bmclaughlin ok mocks say we're gonna 🔪 the rules list in its current form, the free floating cards, so in efforts to save work, suggest we close this pr [these are the latest mocks](https://redhat.invisionapp.com/share/V4NY3C6XKRW#/screens/344290760)

bmclaughlin:
@AllenBW :+1 makes sense.  :cry: for the lost time.

PR #211, opened 2019-02-26
Reconfigure all Actions table filters to configurable (including url)

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHIADVISOR-446

Default sorting of rules is by rule_id ascending, `sort` hates empty params, so we pass the default in to keep it happy

tl;dr you can now clear filters set by the overview page (total_risk and category) this work does not include RHIADVISOR-445 

🌮 💃


bmclaughlin:
Potential follow-up PR idea: avoid user confusion by updating the page header title to match the result set.  Seems outside the scope of this PR.

PR #217, opened 2019-03-04
Abstracts ruletable to component, sets rules/actions lists to consume

AllenBW:
tl;dr - "Rules" tab now navigates to a unfiltered rule table!  yay. and this dry's out a bunch of stuff, more to folllllloooooowwwww

Fixes https://projects.engineering.redhat.com/browse/RHIADVISOR-450

SO this pr begets another, one where we do a crap tonne of cleanup, removing unused components n such, also right now, `ListRules` and `ViewActions` do the same thing (act as a 🏠 for rulestable) and should be named the same, thinking of renaming `ViewActions to `ListActions` and renaming `ListActions to... something that makes sense... cuz its the component for viewing a single rule

### what rules now looks like
screen shot 2019-03-04 at 10 50 32 am

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -1,0 +1,225 @@
+/* eslint camelcase: 0 */
+import React, { Component } from 'react';
+import { Link } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


AllenBW:
cc @Robcwilliams

PR #220, opened 2019-03-05
Update TableToolbar component, maintain pagination right justified

AllenBW:
brought on by  https://github.com/RedHatInsights/insights-frontend-components/pull/343/files

we also updated cleanwebpack plugin, it no longer requires destination to be specified, defaults to output 🎉 

### Gotta maintain that pagination location :yum:
screen shot 2019-03-05 at 3 10 12 pm


ryelo:
@AllenBW two things:
1. It looks like you have a toolbar inside of a toolbar now. The padding is way too much.
2. Can you use the `results` section of the `TableToolbar`? https://github.com/RedHatInsights/insights-frontend-components/blob/master/doc/components/tableToolbar.md#add-results-to-the-toolbar

AllenBW:
@ryelo updated (new picture below!)... and to respond 1. addressed in new picture 2. nope, the filter component already uses pf4 toolbar so now that TableToolbar also uses, yeah its a toolbar inside of a toolbar

SO we're gonna forgo using TableToolbar on the top, instead ` <div className='ins-c-table__toolbar ' style={ { padding: '.1rem' } }>` (i know i know inline style is no bueno)

screen shot 2019-03-06 at 7 24 04 am

ryelo:
@AllenBW 
> SO we're gonna forgo using TableToolbar on the top
I don't think that's a good idea. Considering other apps are using the TableToolbar with the results section mentioned, it's going to look weird to have this table out of place. 

Instead of wrapping `Filters` inside of `TableToolbar`, could you just put `TableToolbar` insider of `Filters`?

Like I said, this is going to be a consistency issue if not implemented the way that other apps are doing it.

AllenBW:
@ryelo consistency issue using pf4 tool bar as specified https://patternfly-react.surge.sh/patternfly-4/demos/toolbar ? 

this is what using `TableToolbar` in the filter component yields
screen shot 2019-03-06 at 7 54 44 am

AllenBW:
Okie dokie, lets give this one a try...

### yah should satisfy, nothing inline
screen shot 2019-03-06 at 8 00 16 am

ryelo:
@AllenBW I'm not sure why there is the empty space in that first screenshots: 

screen shot 2019-03-06 at 8 06 47 am

I was able to repro just fine.

regarding using Patternfly's Toolbar, IFC's TableToolbar is just a shim, it still renders the Patternfly's toolbar under the hood, just with an added className and a section to put the results.

AllenBW:
Oh yeah you wanted that results thing...
screen shot 2019-03-06 at 8 24 55 am

PR #234, opened 2019-03-15
Convert all paths to the cloud.redhat.com paths

AllenBW:
as per https://github.com/RedHatInsights/landing-page-frontend/commit/853ec2f75702baf70591c24ca1c0858b730b9723

note, from browser console run `localStorage.setItem('cs-app', 'insights');`

### so we still have an app/nav mismatch, but @karelhala assures me, latest chrome should fix that (whenver it happens)
Screen Shot 2019-03-18 at 11 27 22 AM

### DIFF
--- a/config/base.webpack.plugins.js
+++ b/config/base.webpack.plugins.js
@@ -84,12 +84,12 @@ const CopyFilesWebpackPlugin = new (require('copy-webpack-plugin'))([
 plugins.push(CopyFilesWebpackPlugin);
 
 /**
- * Replaces any @@insights in the html files with config.insightsDeployment value.
- * This handles the path being either insights or insightsbeta in the esi:include.
+ * Replaces any @@env in the html files with config.deploymentEnv value.
+ * This handles the path being either / or /beta in the esi:include.
  */
 const HtmlReplaceWebpackPlugin = new(require('html-replace-webpack-plugin'))([{
-    pattern: '@@insights',
-    replacement: config.insightsDeployment
+    pattern: '@@env',
+    replacement: config.deploymentEnv
 }]);
 plugins.push(HtmlReplaceWebpackPlugin);
 
@@ -100,10 +100,10 @@ const HotModuleReplacementPlugin = new webpack.HotModuleReplacementPlugin();
 plugins.push(HotModuleReplacementPlugin);
 
 /**
- * Replaces any instance of RELEASE in js files with config.insightsDeployment value.
+ * Replaces any instance of RELEASE in js files with config.release value.
  */
 const Release = new webpack.DefinePlugin({
-    RELEASE: JSON.stringify(config.insightsDeployment)
+    RELEASE: JSON.stringify(config.release)
 });
 plugins.push(Release);
 
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,9 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { routerParams } from '@red-hat-insights/insights-frontend-components';
-import { matchPath } from 'react-router-dom';
 import { connect } from 'react-redux';
+import { matchPath } from 'react-router-dom';
+
 import { Routes } from './Routes';
 import './App.scss';
 
@@ -10,13 +11,11 @@ class App extends Component {
     componentDidMount () {
         insights.chrome.init();
         insights.chrome.identifyApp('advisor');
-        insights.chrome.navigation(buildNavigation());
         this.appNav = insights.chrome.on('APP_NAVIGATION', event => {
-            if (!matchPath(location.href, { path: `${document.baseURI}platform/advisor/${event.navId}` })) {
+            if (!matchPath(location.href, { path: `${document.baseURI}insights/${event.navId}` })) {
                 this.props.history.push(`/${event.navId}`);
             }
         });
-        this.buildNav = this.props.history.listen(() => insights.chrome.navigation(buildNavigation()));
     }
 
     componentWillUnmount () {
@@ -39,17 +38,3 @@ export default routerParams(connect(
     null,
     null
 )(App));
-
-function buildNavigation () {
-    const currentPath = window.location.pathname.split('/').slice(-1)[0];
-    return [{
-        title: 'Actions',
-        id: 'actions'
-    }, {
-        title: 'Rules',
-        id: 'rules'
-    }].map(item => ({
-        ...item,
-        active: item.id === currentPath
-    }));
-}
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -1,4 +1,4 @@
-import { Route, Switch, matchPath } from 'react-router-dom';
+import { matchPath, Route, Switch } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import React from 'react';
 import asyncComponent from './Utilities/asyncComponent';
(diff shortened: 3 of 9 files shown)


dkuc:
This looks right to me.  How about you @iphands

iphands:
@AllenBW no real need to hold off until Monday.
Lemme look... But also feel free to merge away. As we break CI it might be better to test often.
Also lemme get with ya on slack and show how to test this locally.

karelhala:
This is not complete for advisor. With this settings after opening advisor you'd see `cloud.redhat.com/apps/platform/advisor` that is not good. 
You gotta do some other changes you can follow this PR https://github.com/RedHatInsights/insights-frontend-starter-app/pull/62
* [ ] [App.js#L12 ](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/App.js#L12 )  `s/advisor/insights/`
* [ ] [App.js#L43](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/App.js#L43) no longer needed and can be removed.
* [ ] [entry.js#L14](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/entry.js#L14) and [entry-dev.js](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/entry-dev.js) should be updated

It would also be nice to add some default `spandx.config.js` so anybody can start hacking away.

dkuc:
Thanks @karelhala !

iphands:
Also things like https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/Routes.js#L45
Which are not in the Starter App need to be updated.

As well as CSS that might reference assets in `/insights/platform/advisor` or `/insights/static`

proxy/example-local-fronend.js
```diff
@@ -0,0 +1,14 @@
+/*global module, process*/
+
+// Hack so that Mac OSX docker can sub in host.docker.internal instead of localhost
+// see https://docs.docker.com/docker-for-mac/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host
+const localhost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docker.internal';
+
+module.exports = {
+    routes: {
+        '/insights/remediations': { host: `https://${localhost}:8002` },
```
karelhala: `remediations` and `inventory` will be on different port because you'd run them from different webpack.

AllenBW:
@karelhala @iphands  so this works in dev... but clicking on the lnav goes nowhere, BUT IF YOU OPEN IT IN A NEW TAB (success)🤤 🤦‍♂️... wondering if yah have any thoughts? 
Screen Shot 2019-03-18 at 10 06 03 AM

config/base.webpack.plugins.js
```diff
@@ -100,10 +100,10 @@ const HotModuleReplacementPlugin = new webpack.HotModuleReplacementPlugin();
 plugins.push(HotModuleReplacementPlugin);
 
 /**
- * Replaces any instance of RELEASE in js files with config.insightsDeployment value.
+ * Replaces any instance of RELEASE in js files with config.deploymentEnv value.
```
karelhala: Should be `config.deploymentEnv` -> `config.release`

src/Routes.js
```diff
@@ -25,26 +25,24 @@ const paths = {
...
-    return <Route { ...rest } component={ Component } />;
+    return (<Component { ...rest } />);
 };
 
 InsightsRoute.propTypes = {
     component: PropTypes.func,
     rootClass: PropTypes.string
 };
 
-function checkPaths(routes) {
```
karelhala: Do not remove this function, but change it:
```JS
function checkPaths(routes) {
return some(Object
    .values(routes), route => matchPath(location.href, { path: `${document.baseURI}insights${route}` }));
}
```

src/Routes.js
```diff
@@ -53,23 +51,15 @@ function checkPaths(routes) {
  *      path - https://prod.foo.redhat.com:1337/insights/advisor/rules
  *      component - component to be rendered when a route has been chosen.
  */
-export const Routes = ({ childProps: { history }}) => {
-    if (!checkPaths(paths)) {
```
karelhala: You shouldn't remove this part as well, otherwise route links will break

src/entry-dev.js
```diff
@@ -5,18 +5,14 @@ import { Provider } from 'react-redux';
 import { init } from './Store';
 import App from './App';
 import logger from 'redux-logger';
-import { NotificationsPortal } from '@red-hat-insights/insights-frontend-components/components/Notifications';
```
karelhala: Do not remove `NotificationsPortal` othwerwise you will loose notifications.

AllenBW: gonna revert, cuz no reason to do too much here

src/Routes.js
```diff
@@ -53,23 +55,18 @@ function checkPaths(routes) {
...
+export const Routes = (props: Props) => {
+    const path = props.childProps.location.pathname;
+
     if (!checkPaths(paths)) {
         history.push(paths.actions);
     }
 
     return (
         <Switch>
-            <InsightsRoute path={ paths.actions } component={ Actions } rootClass='actions' />
```
karelhala: This shouldn't be changed because we will loose option to bookmark pages.

src/Routes.js
```diff
@@ -53,23 +55,18 @@ function checkPaths(routes) {
  *      path - https://prod.foo.redhat.com:1337/insights/advisor/rules
  *      component - component to be rendered when a route has been chosen.
  */
-export const Routes = ({ childProps: { history }}) => {
+export const Routes = (props: Props) => {
```
karelhala: This shouldn't change, because we are handling redirect in `checkPaths` function.

src/Routes.js
```diff
@@ -29,20 +29,22 @@ const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
...
 InsightsRoute.propTypes = {
     component: PropTypes.func,
     rootClass: PropTypes.string
 };
 
-function checkPaths(routes) {
+function checkPaths (routes) {
     return some(Object
-    .values(routes), route => matchPath(location.href, { path: `${document.baseURI}platform/advisor${route}` }));
+    .values(routes), route => matchPath(location.href, { path: `${document.baseURI}insights/advisor${route}` }));
```
karelhala: This should be just `insights` no `advisor`.

`.values(routes), route => matchPath(location.href, { path: `${document.baseURI}insights${route}` }));`

src/Routes.js
```diff
@@ -29,20 +29,22 @@ const InsightsRoute = ({ component: Component, rootClass, ...rest }) => {
     const root = document.getElementById('root');
     root.removeAttribute('class');
     root.classList.add(`page__${rootClass}`, 'pf-c-page__main');
-    root.classList.add(`page__${rootClass}`, 'pf-l-page__main');
     root.setAttribute('role', 'main');
 
-    return <Route { ...rest } component={ Component } />;
```
karelhala: Do not change this otherwise you'd get incorrect path handling.

src/App.js
```diff
@@ -10,13 +9,6 @@ class App extends Component {
     componentDidMount () {
         insights.chrome.init();
         insights.chrome.identifyApp('advisor');
-        insights.chrome.navigation(buildNavigation());
-        this.appNav = insights.chrome.on('APP_NAVIGATION', event => {
```
karelhala: This shouldn't be removed but updated
```JS
this.appNav = insights.chrome.on('APP_NAVIGATION', event => {
            if (!matchPath(location.href, { path: `${document.baseURI}insights/${event.navId}` })) {
                this.props.history.push(`/${event.navId}`);
            }
        });
```

PR #239, opened 2019-03-20
Update app to support beta branch release as per cloud migration guide

AllenBW:
from https://docs.google.com/document/d/1Ku_MELkCjhY5ggwyQ7aawftiLMoKlqdpdsRPPSr8nyI/edit?ts=5c8fb40a#

### DIFF
--- a/src/entry-dev.js
+++ b/src/entry-dev.js
@@ -7,15 +7,20 @@ import App from './App';
 import logger from 'redux-logger';
 import { NotificationsPortal } from '@red-hat-insights/insights-frontend-components/components/Notifications';
 
-// exposes webpack variable RELEASE
-/*global RELEASE:true*/
+const pathName = window.location.pathname.split('/');
+pathName.shift();
+
+let release = '/';
+if (pathName[0] === 'beta') {
+    release = `/beta/${pathName.shift()}`;
+}
 
 ReactDOM.render(
     <Provider store={ init(logger).getStore() }>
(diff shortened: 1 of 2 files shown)


src/entry-dev.js
```diff
@@ -7,15 +7,20 @@ import App from './App';
...
+
+let release = '/';
+if (pathName[0] === 'beta') {
+    release = `/beta/${pathName.shift()}`;
+}
 
 ReactDOM.render(
     <Provider store={ init(logger).getStore() }>
-        <Router basename={ `${RELEASE ? `/${RELEASE}` : ''}/insights` }>
+        <Router basename={ `${release}${pathName[0]}/${pathName[1]}` }>
```
karelhala: Because you are using only app name you should use just ``${release}${pathName[0]}`.

src/entry.js
```diff
@@ -6,15 +6,20 @@ import { init } from './Store';
...
+
+let release = '/';
+if (pathName[0] === 'beta') {
+    release = `/beta/${pathName.shift()}`;
+}
 
 ReactDOM.render(
     <Provider store={ init().getStore() }>
-        <Router basename={ `${RELEASE ? `/${RELEASE}` : ''}/insights` }>
+        <Router basename={ `${release}${pathName[0]}/${pathName[1]}` }>
```
karelhala: Because you are using only app name you should use just ``${release}${pathName[0]}`.

PR #251, opened 2019-03-25
Removes a few unnecessary divs, adds label to donut legend

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-556

No visual changes to donut 

### Looks a lil like this
Screen Shot 2019-03-25 at 1 44 08 PM

PR #252, opened 2019-03-26
Restores donut outer most div, dashboard doesn't scale well without

### LINKED ISSUE #253: [BUG] Donut hole breaks out of container at larger screen widths
Screen Shot 2019-03-26 at 8 42 56 AM

AllenBW:
slight roll back from #251
fixes #253 

### looks fine for both sizes
Screen Shot 2019-03-26 at 7 10 34 AM
Screen Shot 2019-03-26 at 7 10 22 AM


ryelo:
Related bug: #253

PR #259, opened 2019-03-28
Adds empty state to donut (when totalhits = 0)

AllenBW:
fixes #155
jira https://projects.engineering.redhat.com/browse/RHCLOUD-601

you'll notice the top margin matches the risk severity list
### Looks like this
Screen Shot 2019-03-28 at 2 18 28 PM


AllenBW:
Updated as per @katierik 's comment!

PR #260, opened 2019-03-28
Adds ocp system type for risk of change reporting on rule details page

AllenBW:
### this is what it looks like, system type found and not found
Screen Shot 2019-03-28 at 3 54 24 PM
Screen Shot 2019-03-28 at 3 54 49 PM


AllenBW:
cc @AparnaKarve

PR #262, opened 2019-03-29
Adds routes for access rule and inventory, without category or severity

### LINKED ISSUE #187: add a bookmarkable route without category segment
right now the URL route has the form of
https://access.redhat.com/insightsbeta/platform/advisor/actions/<category>/<rule_id>

If I want to link user to a rule page I may only know its rule_id but not its category. It would therefore be handy if this project also supported routes in form of:

https://access.redhat.com/insightsbeta/platform/advisor/actions/<rule_id>

AllenBW:
fixes #187

so the url: `https://ci.foo.redhat.com:1337/insights/actions/by_id/crashkernel_reservation_failed%7CCRASHKERNEL_RESERVATION_FAILED/`  would take you to the below views, no category or severity necessary

wanted to get @katierik and @robcwilliams thoughts before moving forward with this

Important to note, we don't hyperlink to this anywhere in the app, this would only be for people sending people urls of rules, looking up rules by their ids or w/e

fixes https://projects.engineering.redhat.com/browse/RHCLOUD-394

### First is rule details, second is inventory item from the rule details
Screen Shot 2019-03-29 at 12 59 39 PM
Screen Shot 2019-03-29 at 12 59 50 PM

cc @jharting


jharting:
@AllenBW works for me. Alternatively, once rule data is fetched (and you get to know the category) you may want to redirect to `/actions/:type/:id`. Not sure if it's worth it, though

Robcwilliams:
Good with the change. Re: redirect - let's proceed as is for now and we can always add it later depending on how we see this endpoint being used.

PR #269, opened 2019-04-03
Resync with starter application Cloud fixes/refactor

iphands:
The biggest issue here was the esi:include was looking for src="apps/chrome...
This needs to be src="/apps/chrome...

Other refactoring changes pulled in as well


AllenBW:
and we have a fix! gonna wait for https://github.com/RedHatInsights/insights-chrome/pull/232 till testing this...

PR #286, opened 2019-04-15
Grabbag of ux changes

AllenBW:
tl;dr: first ss is the most latest, has all changes, the rest just show the change indicated in the title... this breaks sorting by "Added" will work as soon as api is updated

### we also updated the ansible row (as per mocks)
Screen Shot 2019-04-15 at 11 31 48 AM

### We updated the table as per mocks, results in upper right, different rows, relative time, search icon in search bar
Screen Shot 2019-04-15 at 10 12 06 AM

### Move "show rule with hits" to the left of the toolbar
Screen Shot 2019-04-15 at 8 23 58 AM

### inactive to disabled
Screen Shot 2019-04-15 at 8 28 09 AM

### actions only happen on root row
Screen Shot 2019-04-15 at 8 40 01 AM


AllenBW:
cc @robcwilliams

Robcwilliams:
+1

PR #289, opened 2019-04-16
Just a wip for the sake of wipping

AllenBW:
### bummer, no results, looks like we're only showing enabled rules, lets change that!
Screen Shot 2019-04-17 at 11 02 21 AM

### dang still no hits, oh well, at least we see that we went from filtering by only enabled to all rules! neato! 🎉  and look at that nice updated message!
Screen Shot 2019-04-17 at 11 41 43 AM

~We look finish, but need finish two pieces before this is g2g~
~* get the filtering interaction down solid (when clicking the link)~
~* center the message state, this is tricky cuz we are an action table~

### DIFF
--- a/src/PresentationalComponents/Filters/Filters.js
+++ b/src/PresentationalComponents/Filters/Filters.js
@@ -17,10 +17,11 @@ class Filters extends Component {
 
     componentDidUpdate (prevProps) {
         if (this.props.externalFilters !== prevProps.externalFilters) {
-            const filterKey = Object.keys(this.props.externalFilters)[0];
-            if (filterKey) {
-                this.props.setFilters({ [filterKey]: `${this.props.externalFilters[filterKey]}` });
-            }
+            let filters = {};
+            Object.entries(this.props.externalFilters).forEach(
+                ([ key, value ]) =>  filters[key] = value
+            );
+            this.props.setFilters(filters);
         }
     }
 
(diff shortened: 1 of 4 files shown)


AllenBW:
cc @robcwilliams

katierik:
Woohoo!
I realize that I didn't provide an empty state design for when the disabled rules are included. Here is that: https://redhat.invisionapp.com/share/V4NY3C6XKRW#/358716417_Empty_State_-_No_Rule_Hits_W_Disabled

Also tiny typo, should be "enabled" rules, not "enable" in the message

AllenBW:
Oooo fixed typo, the other thing is gonna take a hot second or two 😢

src/PresentationalComponents/Filters/Filters.js
```diff
@@ -17,10 +17,11 @@ class Filters extends Component {
 
     componentDidUpdate (prevProps) {
         if (this.props.externalFilters !== prevProps.externalFilters) {
-            const filterKey = Object.keys(this.props.externalFilters)[0];
-            if (filterKey) {
-                this.props.setFilters({ [filterKey]: `${this.props.externalFilters[filterKey]}` });
-            }
+            let filters = {};
+            Object.entries(this.props.externalFilters).forEach(
```
dkuc: Looks like your just copying externalFilters into a new object called filters?

why not `let filters = {...this.props.externalFilters}'`

AllenBW: woah, good call, we can simplify this quite a bit, sorry got a lil crazy there...

dkuc:
Approved

PR #302, opened 2019-04-25
As per ux design, going with large table rather than compact

AllenBW:
fixes https://docs.google.com/spreadsheets/d/1xbzvLhRUd0o27CIk1MOQWbpYMpJXktx3HUpI2MAjpk4/edit?ts=5cc0cc87#gid=967324608

### tada, bigger, n supports expanded table nicely
Screen Shot 2019-04-25 at 12 18 45 PM

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -95,7 +95,7 @@ class RulesTable extends Component {
                                 <div key={ key }>
                                     { moment(value.created_at).fromNow() }
                                 </div>,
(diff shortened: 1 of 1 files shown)


katierik:
Awesome!

Two small notes on this page:
The expanded table is missing some spacers in it. I think the lg pf spacer is correct here.
Xnip2019-04-25_11-37-00

The items in the table should all be top-aligned. The risk icon is too high.
Xnip2019-04-25_11-38-08

AllenBW:
Updated original ss as per feedback @katierik  now with more child spacing and top vertical alignment for battery!

PR #307, opened 2019-04-30
Moves rule details inventory table title outside of card

AllenBW:
### what it looks likeeeee, inline with mocks
Screen Shot 2019-04-30 at 11 00 41 AM


AllenBW:
@katierik updated as per feedback!

PR #310, opened 2019-05-02
Add pagination function to affected systems

karelhala:
There's been a change to how inventory is fetching next pages, it used to handle pagination for ourselves, but it has changed in more robust way and now we have to paginate on our side.

In memory as of for now, we should call get current page on each `onRefresh` call so we get newest data.


AllenBW:
@karelhala not seeing any systems now 😬 
Screen Shot 2019-05-02 at 6 43 54 AM

PR #314, opened 2019-05-06
Pass page size down to inventory so we have correct pagination in place

karelhala:


AllenBW:
confirmed okie dokie working on beta, pushing to stable ❤️ thanks again @karelhala !

PR #319, opened 2019-05-14
Use new frontend components monrepo packages

karelhala:
This is how we should start using FEC, they are moving to brand new repo - https://github.com/RedHatInsights/frontend-components


AllenBW:
Oh this is glorious glorious glorious.

AllenBW:
Looks like we're missing the remediate button on the inventory table... looking into this...

PR #334, opened 2019-06-03
Update overview donut to better align with pf styles (black tooltip)

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1433

[pf friends updated the donut to be better, so we'll update the donut to be better](http://patternfly-react.surge.sh/patternfly-4/charts/chartdonut/)

**Notes**
* all behavior, aside from mouseover slice highlight, is preserved (but that wasn't ever a pf thing to begin with)
* satisfies https://redhat.invisionapp.com/share/V4NY3C6XKRW#/357403104_Donut_Interactions
* uses bunches of less code 🤤 

### What it looks...
Screen Shot 2019-06-03 at 8 33 14 AM
Screen Shot 2019-06-03 at 9 13 09 AM

### Now when we zoom in, it maintains sanity!! (previously it would freak out)
Screen Shot 2019-06-03 at 8 33 41 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1487,15 +1487,38 @@
       "integrity": "sha512-x+zxOCEOtxqoLwBQK+I2BB1CGY/uX40uFpE4RMUPv75W8BLegC0jvXHMM+BK5eP3680AMfVZn5xbfQVAVzyT1A=="
     },
     "@patternfly/react-charts": {
-      "version": "3.4.2",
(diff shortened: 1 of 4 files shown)


AllenBW:
cc @katierik for a quick 👀 
cc @robcwilliams for awareness

katierik:
Looks great!

PR #338, opened 2019-06-06
Removes unnecessary deps/utils

AllenBW:
The goal here is to remove everything we dont need (deps/configs/utils) n keep all the thing we do need to keep it running... a little bit of house cleaning if yah will, and relying on `@redhat-cloud-services/frontend-components-config`

Oh and all the build commands still work just fine...

Loosely based on https://github.com/RedHatInsights/insights-frontend-starter-app/pull/71/

before this can move forward... we need: 
* ~https://github.com/RedHatInsights/frontend-components/pull/43~
* i lied, https://github.com/RedHatInsights/frontend-components/pull/45 is what we really need, and now we have it

### like nothing ever happened
Screen Shot 2019-06-07 at 6 52 10 AM


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "insights-advisor-frontend",
-  "version": "0.2.0",
+  "version": "1.0.0",
```
ryelo: It's a big day!

package.json
```diff
@@ -57,76 +58,44 @@
...
+    "webpack-dev-server": "3.7.0"
   },
   "scripts": {
     "build": "webpack --config config/prod.webpack.config.js",
     "test": "jest --passWithNoTests",
     "lint": "npm-run-all lint:*",
     "lint:js": "eslint config src",
     "lint:sass": "stylelint 'src/**/*.scss' --config .stylelintrc.json",
-    "prod": "NODE_ENV=production webpack-dev-server",
+    "prod": "NODE_ENV=production webpack-dev-server --config config/dev.webpack.config.js",
```
dkuc: prod and dev use the same webpack config?

karelhala: That is prod start, so the files are minified and you can check live reloads as if app was deployed. he actual build happens in `build` script

AllenBW: yep! the prod and dev webpack config are exactly the same, could maybe get rid of one of em for giggles? 🤷‍♂ (but secretly i don't want to dismiss all this great review)

ryelo: It's essentially emulating prod, but locally. This doesn't actually build what will be live on public prod.

PR #382, opened 2019-07-30
Api calls need to do an auth check, this consolidates em in one place

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-1953
makes it so we don't have to worry about this for any future calls

### still not broken
Screen Shot 2019-07-30 at 1 42 53 PM

### DIFF
--- a/src/AppActions.js
+++ b/src/AppActions.js
@@ -1,7 +1,11 @@
 import * as ActionTypes from './AppConstants';
 import API from './Utilities/Api';
 
-const fetchData = async (url, headers, options) => (await API.get(url, headers, options)).data;
+const fetchData = async (url, headers, options) => {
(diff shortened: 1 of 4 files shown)


src/AppActions.js
```diff
@@ -1,7 +1,11 @@
 import * as ActionTypes from './AppConstants';
 import API from './Utilities/Api';
 
-const fetchData = async (url, headers, options) => (await API.get(url, headers, options)).data;
+const fetchData = async (url, headers, options) => {
+    await insights.chrome.auth.getUser();
```
dkuc: So most of the time, this would be a no-op right?

AllenBW: yup thats, what Ive been told 🌮

PR #383, opened 2019-07-31
Add script to easily startup Insights API environment

mhuth:


mhuth:
@dkuc you mentioned this in the old PR:
"As part of the script, you may need to edit config/spandx.config.js to get insights-proxy to talk to this."

Can you give me an idea of how this script might need to be changed.  I've been testing with different ports for the API and it all seems to still work with the frontend without modifying that config.  Maybe even a demo might help me.

dkuc:
@mhuth This actually took me a while toi figure out but I finally got my local front end to run against my local api with this spanx config.

```
const apiHost = `http://127.0.0.1:8000`;
const frontendHost = (process.env.PLATFORM === 'linux') ? 'localhost' : 'host.docker.internal';
const chromeHost = 'https://ci.cloud.paas.psi.redhat.com';

exports.routes = {
    '/api/insights/v1/': { host: apiHost },
    '/apps/chrome': { host: chromeHost },
    '/apps/insights/': { host: `http://${frontendHost}:8002` },
    '/insights': { host: `http://${frontendHost}:8002` }
};
```

dkuc:
I created #387 so that we can start from a more reasonable default.

dkuc:
@mhuth #387 is now merged.  Let me know if you run into any problems.

mhuth:
@dkuc I've updated the script to account for the new spandx config.  Thanks for that.  I use your docker command to start the insights-proxy container in the script.  I had to change the docker command slightly to make it work for me.  I had to use `docker run -v $PWD/config/spandx.config.js:/config/spandx.config.js ...`

dkuc:
I got it to work! nice.
@mhuth We need a way to seed the database so it has some data in it.

PR #393, opened 2019-08-07
Rulestable on init informs state filter( is updated when changes occur)

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-410

### WHAT THIS DOES?!
When you throw a url like `https://ci.foo.redhat.com:1337/insights/rules?impacting=false&reports_shown=true&total_risk=4,3` into the browser (already logged in) you go to a page that looks like this
Screen Shot 2019-08-07 at 3 58 04 PM

as yah add more filters the url is updated to reflect the selected filters

navigating forwards/backwards/sidewards throughout the app doesn't do anything hokey

also, if you visit `/rules` with NO filters, whatever is in the filters state is pushed in the url

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -26,7 +26,7 @@ import MessageState from '../MessageState/MessageState';
 import RuleDetails from '../RuleDetails/RuleDetails';
 
 const RulesTable = (props) => {
-    const { rules, filters, rulesFetchStatus, setFilters, fetchRules, addNotification } = props;
+    const { rules, filters, rulesFetchStatus, setFilters, fetchRules, addNotification, history } = props;
     const [cols] = useState([
         { title: 'Rule', transforms: [sortable] },
         { title: 'Added', transforms: [sortable, cellWidth(15)] },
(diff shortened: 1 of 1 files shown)


AllenBW:
@karelhala @ryelo if yah have the time to spare a quick sanity check, would appreciate it on this one as its kinda tricky

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -126,14 +128,37 @@ const RulesTable = (props) => {
...
 
     useEffect(() => {
-        fetchRules({
-            ...filters,
-            offset,
-            limit,
-            impacting,
-            sort
+        if (history.location.search) {
+            const searchParams = new URLSearchParams(history.location.search);
```
karelhala: I think that this could be simplified with reduce function right? Also it might be better to use `JSON.parse` to get real value from params.
```JS
const searchParams = new URLSearchParams(history.location.search);
const paramsObject = Array.from(searchParams).reduce((acc, [key, value]) => ({
  ...acc,
  [key]: JSON.parse(value)
}), {});
setImpacting(paramsObject.impacting);
setFilters({ ...paramsObject });
```

AllenBW: this is awesome! ahhhhh had a feeling reduce coulda been used but couldn't quite get there in the heat of the moment, Json.parse won't quite work for the number strings (which gotta stay strings) but thats a great solution for pulling out the booleans!

PR #398, opened 2019-08-19
Reconfigures topic filter to be set on card rather than in component

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2106


AllenBW:
Hopefully this makes @kensible day! 🤞

PR #399, opened 2019-08-19
Adds system view to Insights

AllenBW:
### what it presently looks like, url updates while tab clickin
Screen Shot 2019-08-20 at 1 03 22 PM
Screen Shot 2019-10-07 at 4 55 01 PM

#### and then when you click a system you see this (breadcrumbs yo)
Screen Shot 2019-10-08 at 8 15 16 AM


katierik:
For the filtered/searched empty state, we should use this one: https://marvelapp.com/i4441ie/screen/58398632 (not the happy check)

Number of rule hits is a link in the mocks - should open the system view with the rule hits showing.

Breadcrumb should be: "Rules -> Systems -> [System]" (wasn't in mocks, was a standardizing decision after that - updating them now)

AllenBW:
> Number of rule hits is a link in the mocks - should open the system view with the rule hits showing. 

☝️ i don't follow, how is that different from clicking on the system name? 

> Breadcrumb should be: "Rules -> Systems -> [System]" (wasn't in mocks, was a standardizing decision after that - updating them now) --> Actually... We have "Rules" as the nav and "Rules" as the tab so if it was a rule it would be "Rules -> Rules -> [Rule name]" which is ok I guess.

should the url follow this pattern?  instead of /systems it should say /rules/systems/?

AllenBW:
@katierik ☝️

AllenBW:
SO we'll be
* updating the breadcrumbs in another pr, cuz it needs to be done for both rules and systems
* checking with rob if he wants the system number to also link (to the same place as the display name)

but that can be modified in another pr as well so 🤷‍♂ :godmode: ⏩

PR #403, opened 2019-08-26
Adds support for translation

AllenBW:
**this looks big, but thats a lie** its just a lotta of the same everywhere in the app, adds support for translations everywhere except the [constants file](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/AppConstants.js) but that aside, everywhere else is taken care of

fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2164


AllenBW:
Ok still gotta figure out how to add translation support to our[constants file](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/AppConstants.js) but that aside, everywhere else is taken care of...

AllenBW:
@ryelo @karelhala if ya'll can pop in a give this a quick glace... OR throw ideas out how to do the constants file... i would... 🙇‍♀ ❤️ 💃 🌮

PR #410, opened 2019-09-05
Show disabled remediate with ansible button

AllenBW:
to align with mocks here https://marvelapp.com/i4441ie/screen/58398647


AllenBW:
gonna close, only gonna update the button and use the ansible icon everywhere else we use that hand rolled svg (because it now exists)

PR #423, opened 2019-09-19
Edit api-setup script for mac

LI0131:


scripts/setup_insights_api.sh
```diff
@@ -138,7 +138,7 @@ fi) &
 
 echo_bold "Starting insights-proxy container  ..."
 sed "s|apiHost = .*$|apiHost = \`http://localhost:${API_PORT}\`;|" $REPO_SPANDX_CONFIG > $TMP_SPANDX_CONFIG
-docker run -d --name $PROXY_CONTAINER --net='host' -p $PROXY_PORT:1337 \
```
mhuth: Damn, is it really not possible to put the proxy container on the host network?  It needs to communicate with 127.0.0.1:8002 on the host and I'm not sure how to do this if the proxy is running on another container network.  Any ideas?

LI0131: I can look more into it tomorrow. I wasn't able to find any other way to connect to the proxy other than taking the proxy container off the host network. It might be possible though.

PR #427, opened 2019-09-23
Adds bulk select to rule details inventory table (bulk bulk bulk bulk)

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2069

#### To boot what we see
Screen Shot 2019-09-24 at 10 57 00 AM

#### there are a few options
Screen Shot 2019-09-24 at 10 58 47 AM

#### selectin em all lights all checkboxes up, and changes bulk checkbox value
Screen Shot 2019-09-24 at 10 57 28 AM

#### further proceeding with remediation shows we caught em all
Screen Shot 2019-09-24 at 10 57 39 AM

#### but also it still works if you wanta pick only a few
Screen Shot 2019-09-24 at 10 59 12 AM

#### if you've no playbooks nothing is shown
Screen Shot 2019-09-24 at 10 59 49 AM

### DIFF
--- a/locales/data.json
+++ b/locales/data.json
@@ -61,6 +61,9 @@
     "title.topics": "Topics",
     "topicslist.notopics.title": "There was an issue fetching topics",
     "topicslist.notopics.body": "Either no topics presently exist or there is an issue presenting them.",
-    "remediate": "Remediate"
+    "remediate": "Remediate",
+    "selectAll": "Select all ({items} items)",
+    "selectPage": "Select page ({items} items)",
+    "selectNone": "Select none (0 items)"
   }
 }
--- a/locales/translations.json
+++ b/locales/translations.json
@@ -60,5 +60,8 @@
   "title.topics": "Topics",
   "topicslist.notopics.title": "There was an issue fetching topics",
   "topicslist.notopics.body": "Either no topics presently exist or there is an issue presenting them.",
-  "remediate": "Remediate"
+  "remediate": "Remediate",
+  "selectAll": "Select all ({items} items)",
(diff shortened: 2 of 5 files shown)


AllenBW:
@karelhala there's only one rub, that is, looks like checkbox value doesn't update when `checked` prop changes (via value rather than click)... might need need to throw that in a `didupdate` 🤔

AllenBW:
@katierik wondering if the attached ss is the latest, if this work should be tweaked accordingly? 
Screen Shot 2019-09-24 at 11 17 57 AM

katierik:
Yea - the column header level select should be removed (so only the row-level and the toolbar-level). "Select page" should be in there as well.

AllenBW:
@katierik roger that! we pass in `canSelectAll={false}` to the inv table so as soon as it supports consumer props, the checkall will disappear

AllenBW:
#### Added page option!
Screen Shot 2019-09-25 at 7 44 20 AM

AllenBW:
We'll also want to bring in this work https://github.com/RedHatInsights/frontend-components/pull/159

AllenBW:
OK so we're all good for checkbox to disappear, which will happen when https://github.com/RedHatInsights/insights-chrome/pull/365 is merged (and then progress through the chrome build envs, ci - qa - prod beta/stable)

SO we can merge this now and expect to see the checkbox vanish by weeks end

AllenBW:
cc @imtayadeway and @jntullo for bulk input

jntullo:
What beautiful bulk! 😻

PR #444, opened 2019-10-07
Update rule detail topic card as per ux feedback

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2654

#### looks like this!
Screen Shot 2019-10-07 at 1 09 56 PM


katierik:
Color looks good. Still has a shadow which we don't want
Xnip2019-10-07_13-21-39

AllenBW:
Right on! box shadow no more
Screen Shot 2019-10-07 at 1 26 05 PM

PR #453, opened 2019-10-15
Address ux feedback for rules details component

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2750

We don't manage inv table, so the empty state wasn't touched, all other feedback in the issue has been addressed (as of this pr)

### DIFF
--- a/src/PresentationalComponents/RuleDetails/_RuleDetails.scss
+++ b/src/PresentationalComponents/RuleDetails/_RuleDetails.scss
@@ -32,3 +32,11 @@
   background-color: var(--pf-global--BackgroundColor--150);
   box-shadow: none;
 }
+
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/RuleDetails/_RuleDetails.scss
```diff
@@ -31,4 +31,12 @@
...
   background-color: var(--pf-global--BackgroundColor--150);
   box-shadow: none;
+}
+
+.alignCenterOverride {
+  align-items: center;
+}
+
+.batteryTextMarginOverride {
+  padding-top: 1px;
```
AllenBW: i know i know 😭 but there truly is no other way 😭

PR #461, opened 2019-10-21
Adds filtered export support to rules table

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2607

`/api/insights/v1/export/hits.json/?total_risk=4&reports_shown=true&impacting=true`

~~returns slightly more (20) than 1 system (which is what we see in the rules table)~~ i am told this is expected

SILVER LINING all systems have `"total_risk":4` which is **AWESOME** 🙇‍♀ 🤗 

text filtering doesn't yet work, but that wont change anything here, so we can go ahead n merge thissss i believe


AllenBW:
@robcwilliams you ok to merge this?

PR #462, opened 2019-10-22
Updates rules table toolbar to use FE primary toolbar component

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2669
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2754

**TODO**

- filters
- correct strings
- everything else we forgot

### wip looks like this
Screen Shot 2019-10-22 at 10 15 38 AM


src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -64,21 +64,23 @@ const RulesTable = ({ rules, filters, rulesFetchStatus, setFilters, fetchRules,
...
     const toggleRulesWithHits = (impacting) => {
         setFilters({ ...filters, impacting });
         setImpacting(impacting);
         setOffset(0);
     };
 
+    const toggleRulesDisabled = (reports_shown) => {
+        setFilters({ ...filters, reports_shown });
+        setReportsShown(reports_shown);
+        setOffset(0);
```
Hyperkid123: I would consider using `useReducer` for this component. You are doing batch updates which are not playing well with `useState`. This callback will do one re-render per setState call. Using `useReducer` you can bundle this to one dispatch which do only one state update and only one render. Probably not in this PR but something that might be good to think about.

AllenBW: oooo https://reactjs.org/docs/hooks-reference.html#usereducer thanks for the tip @Hyperkid123 🤗

PR #469, opened 2019-11-01
Refactors app.js to be functional component

AllenBW:
sort of almost fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2926

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,59 +1,46 @@
-import PropTypes from 'prop-types';
-import React, { Component } from 'react';
-import routerParams from '@redhat-cloud-services/frontend-components-utilities/files/RouterParams';
-import { connect } from 'react-redux';
-import { matchPath } from 'react-router-dom';
-
-import { Routes } from './Routes';
 import './App.scss';
 
-class App extends Component {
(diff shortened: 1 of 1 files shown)


AllenBW:
SOOO we're gonna hold off on mergin this till https://github.com/RedHatInsights/insights-chrome/pull/434 is in... it oughta fix the issue platform-experience side... we still wanta update this piece, but can hold off till then

src/App.js
```diff
@@ -1,59 +1,46 @@
...
-    render() {
-        return (
-            <Routes childProps={this.props} />
-        );
-    }
-}
+    useEffect(() => {
+        const baseComponentUrl = props.location.pathname.split('/')[1];
+        appNavClick[baseComponentUrl](false);
+    });
```
karelhala: I don't like this one. Can you update it to `useEffect` with some props watch? Like, when something from router updates? Because you are running this on every update and it can lead to infinit loop.

AllenBW: OOOO yeaaaaaah lets add both appnavclick and props.location to that... 🤔  testing now...

PR #477, opened 2019-11-13
Adds disable rule justification modal and action to rule detail page

AllenBW:
**TODO:**
* ~api call n such for single system and justification~ https://projects.engineering.redhat.com/browse/RHCLOUD-2986 is where this will be done, but we laid the groundwork here
* ~possibly rolling https://projects.engineering.redhat.com/browse/RHCLOUD-2982  into this one (cuz it would be kinda ah jerk move to allow disabling on rule details without showing it)~ given api toggle is available (enable disable) this can remain separate  

### as per ux guidance, details page now looks like this, i didn't want to updated the gif
Screen Shot 2019-11-13 at 4 04 06 PM

### Just in case you have some 🍿 lying around
disable

fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2979

### DIFF
--- a/src/PresentationalComponents/Modals/DisableRule.js
+++ b/src/PresentationalComponents/Modals/DisableRule.js
@@ -1,0 +1,105 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Modal, Button, Form, FormGroup, TextInput, Checkbox } from '@patternfly/react-core';
+import { injectIntl } from 'react-intl';
+import { connect } from 'react-redux';
+import { addNotification } from '@redhat-cloud-services/frontend-components-notifications';
+
+import API from '../../Utilities/Api';
+import { BASE_URL } from '../../AppConstants';
+import messages from '../../Messages';
+
+const DisableRule = ({ handleModalToggle, intl, isModalOpen, displaySingleSystemCheckbox, rule, afterDisableFn, addNotification }) => {
+    const [justification, setJustificaton] = useState('');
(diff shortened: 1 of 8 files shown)


katierik:
The field looks good. Missing showing the justification note (mock: https://marvelapp.com/i4441ie/screen/62092193 )
Actions is in the wrong spot too. Note where it is in the Compliance app:
Xnip2019-11-13_14-48-10

AllenBW:
@katierik good call on actions position, as for the justification note, as mentioned above, that'll be done in another pr

src/PresentationalComponents/Modals/DisableRule.js
```diff
@@ -0,0 +1,105 @@
...
+
+const DisableRule = ({ handleModalToggle, intl, isModalOpen, displaySingleSystemCheckbox, rule, afterDisableFn, addNotification }) => {
+    const [justification, setJustificaton] = useState('');
+    const [singleSystem, setSingleSystem] = useState(false);
+
+    const disableRule = async () => {
+        try {
+            if (rule.reports_shown) {
+                // eslint-disable-next-line camelcase
+                await API.post(`${BASE_URL}/ack/`, { rule_id: rule.rule_id, justification });
```
Hyperkid123: I am just curious. Is the trailing `/` required in the URL?

AllenBW: yeah, otherwise we get a 301 :-/

PR #479, opened 2019-11-14
Add disable rule message state to rule detail component

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2982

#### looks like
Screen Shot 2019-11-15 at 8 37 26 AM
Screen Shot 2019-11-15 at 3 13 58 PM

### DIFF
--- a/locales/data.json
+++ b/locales/data.json
@@ -11,6 +11,9 @@
     "disableRule": "Disable rule",
     "disableRuleBody": "Disabling a rule means that hits for this rule across all systems will not be shown in reports and dashboards.",
     "disableRuleSingleSystem": "Disable only for this system",
+    "ruleIsDisabled": "Rule is disabled",
+    "ruleIsDisabledBody": "This rule has been disabled and has no results.",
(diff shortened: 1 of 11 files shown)


katierik:
1 - Is that the empty state format from PF? (I honestly can't tell, but the bell looks big and a little dark so I ask)
2 - For the message about it being disabled, should use the default inline alert style from pf: https://www.patternfly.org/v4/documentation/react/components/alert (which is not the color in the mock argh sorry)
3 - Typo on "following" (says "folling")
4 - The bell icon in the disabled label doesn't have any padding around the top& bottom. Its 2px top & bottom in the mocks - maybe making the icon smaller?

AllenBW:
@katierik 

1 - yes using md bell
2 - which message should use this? the one in the card? 
3 - thats just an older ss, code does not reflect this
4 - bell icon made smaller looks like this
Screen Shot 2019-11-15 at 2 57 01 PM

AllenBW:
we can use the global info color 200 to make thissssss (an alert thats not an alert)
Screen Shot 2019-11-18 at 6 40 25 AM

PR #483, opened 2019-11-21
Add Topics Admin Table

LI0131:
This PR is in response to RHCLOUD-2988. This will add a new Topics Admin table for internal users, and will include a Modal for that page to create new Topics.


LI0131:
@AllenBW I need some guidance on getting the `Edit Topic` button to remain right justified during table resizing, but beyond that, this PR _should_ be RFR

AllenBW:
@LI0131 does that mean we still need api stuff? orrrrrr yer workin that?

AllenBW:
Oooo real quick would yah mind installin https://marketplace.visualstudio.com/items?itemName=amatiasq.sort-imports n letting it do its thing?

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+
+    const setDeleteHandler = ({ type, options }) => {
+        type === 'DELETE' ?
+            deleteTopic(options) :
+            setTopic(options);
+        handleModalToggle(false);
+    };
+
+    const footer = (
+        <Split gutter='sm' style={{ width: '100%' }}>
```
AllenBW: Lets keep style in the .scss 👍

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+    const prevTopic = usePrevious(topic);
+
+    useEffect(() => {
+        if (topic !== prevTopic) {
+            insertTopicValues();
+        }
+    });
+
+    return <Modal
+        width={'60%'}
```
AllenBW: style in da .scss

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+                    name='topic-form-description'
+                    id='topic-form-description'
+                    onChange={ (desc) => setDescription(desc) }
+                />
+            </FormGroup>
+            <FormGroup isInline fieldId='topic-form-labels'>
+                <FormGroup
+                    label='Label'
+                    fieldId='topic-form-tag'
+                    helperText='Rules tagged with this tag will be added to the topic'
```
AllenBW: If this string might require translation, off to messages.js!

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,196 @@
...
+            </Title>
+        </PageHeader>
+        <Main>
+            <React.Fragment>
+                <PrimaryToolbar className='toolbar-padding-override'>
+                    <Button
+                        variant='primary'
+                        onClick={ (_) => hideTopics(_) }
+                    >
+                        Create new topic
```
AllenBW: looks like a string, BETTER TRANSLATE IT 🙃 🛴

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,196 @@
...
+
+    return <React.Fragment>
+        {addEditTopicOpen && <AddEditTopic
+            isModalOpen={addEditTopicOpen}
+            handleModalToggle={setAddEditTopicOpen}
+            topic={selectedTopic}
+        />}
+        <PageHeader>
+            <Title headingLevel="h3" size="2xl" className='titlePaddingOverride'>
+                Topics - administration
```
AllenBW: STRINNNNNNNGGGGGG

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,196 @@
...
+                        </MessageState>),
+                    props: { colSpan: 7 }
+                }]
+            }]);
+        } else {
+            const rows = topicsArray.flatMap((value, key) => ([{
+                isOpen: false,
+                rule: value,
+                cells: [{
+                    title: <div key={key}> {value.name}</div>
```
AllenBW: can we make these `<span>`s ?

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,196 @@
...
+        setTopicsArray(
+            Immutable.from(
+                topicsArray.asMutable().sort((a, b) =>
+                    a[index] > b[index] ? 1 : -1
+                )
+            )
+        );
+    }, [setSortBy, setTopicsArray, topicsArray]);
+
+    const hideTopics = async (topic) => {
```
AllenBW: this doesn't need to be async or the try catch block (also could kill the notification)

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+                    isChecked={ enabled ? true : false }
+                    id='enabled'
+                    label='Enabled (visible)'
+                    onChange={ () => setEnabled(!enabled) }
+                    className='radio-override'
+                />
+            </FormGroup>
+            <FormGroup label='Featured' fieldId='topic-form-featured'>
+                <Checkbox
+                    isChecked={ featured ? true : false}
```
AllenBW: this is already a bool (i believe) could rewrite as `isChecked={ featured }` ?

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+                        name='topic-form-name-2'
+                        onChange={ (name) => setName(name) }
+                    />
+                </FormGroup>
+            </FormGroup>
+            <FormGroup label='Status' fieldId='topic-form-enabled'>
+                <Radio
+                    isChecked={ !enabled ? true : false }
+                    id='disabled'
+                    label='Disabled (not visible)'
```
AllenBW: Hi string. Make me translatable. K thanks bye.

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+        type === 'DELETE' ?
+            deleteTopic(options) :
+            setTopic(options);
+        handleModalToggle(false);
+    };
+
+    const footer = (
+        <Split gutter='sm' style={{ width: '100%' }}>
+            <SplitItem>
+                <Button key='confirm' variant='primary' onClick={ () => setDeleteHandler({
```
AllenBW: yah do this twice, `setDeleteHandler` maybe abstract as a function somewhere else and pass in the type?

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,193 @@
...
+                <Button key='delete' variant='danger' onClick={ () => setDeleteHandler({
+                    type: 'DELETE', options: { slug: tag, name, description, enabled, featured }
+                })}>
+                    {intl.formatMessage(messages.deleteTopic)}
+                </Button>
+            </SplitItem>
+        </Split>
+    );
+
+    const insertTopicValues = () => {
```
AllenBW: from here to 76, can yah walk me through what we're doing here?  (i :suspect:  we can greatly simplify da task ) (maybe over 👖 )

LI0131:
@AllenBW After looking into the API stuff further it looks like DELETE and POST are actually being inherited through Django (which is why they aren't in the swagger spec). So we should be good to go on that! I'll start working through the changes you requested 👍

src/PresentationalComponents/Modals/AddEditTopic.js
LI0131: So I can remove the `insertTopicValues` and move this to the `useState` for each value and then use the setter methods for each in the effect on line 72. For the other two methods, I was using these to compare the previous state to the current state so I know when to update the values of the topic, which then updates the values that are actually displayed. I wasn't 100% sure how to make that work, so you probably will have a better solution 😆

AllenBW: dayum. late worker, they must be payin yah over time 😏  we'll do ah 👖 tomorrow! whenever yer around n talk about it 💃

AllenBW:
Jumping in, a few more pieces, 

*  lets kill row kebab menu
* when editing a  topic, status doesn't populate, maybe need to cast from string to bool? 
* something i need to look into... gettin a 403 on any actual action, which is GOOD cuz, we don't want peeps makin em willy nilly, but we gotta test it... n on ci/qa we oughta be able to create/edit/delete 🤔

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,185 @@
...
+import { connect } from 'react-redux';
+import { injectIntl } from 'react-intl';
+import messages from '../../Messages';
+
+const AddEditTopic = ({ handleModalToggle, intl, isModalOpen, topic, setTopic, deleteTopic }) => {
+
+    const [name, setName] = useState(topic.name || '');
+    const [description, setDescription] = useState(topic.description || '');
+    const [tag, setTag] = useState(topic.tag || '');
+    const [enabled, setEnabled] = useState(false);
```
AllenBW: topic.enabled?

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,188 @@
...
+                cells: [{
+                    title: <span key={key}> {value.name}</span>
+                }, {
+                    title: <span key={key}> {value.tags}</span>
+                }, {
+                    title: <span key={key}> {value.slug}</span>
+                }, {
+                    title: <span>
+                        {value.enabled ? <span>
+                            <CheckCircleIcon color='#92d400'/> {intl.formatMessage(messages.enabled)}
```
AllenBW: lets do the color as a class, set it using pf-success variable in the .scss

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,188 @@
...
+                        </span> : <span>
+                            <BanIcon color='gray'/> {intl.formatMessage(messages.disabled)}
```
AllenBW: same thing as above, class in the scss

src/PresentationalComponents/Modals/AddEditTopic.js
```diff
@@ -0,0 +1,185 @@
...
+                <FormGroup
+                    label={intl.formatMessage(messages.label)}
+                    fieldId='topic-form-tag'
+                    helperText={intl.formatMessage(
+                        messages.topicAddEditTagHelperText
+                    )}
+                    className='text-input-override'
+                >
+                    <TextInput
+                        value={tag}
```
AllenBW: maybe do `value={tag.replace(/\s/g,'').toLowerCase()}`

and on 114 ` onChange={ (tag) => setTag(tag.replace(/\s/g,'').toLowerCase()) }`

mocks show label as being without whitespace, gonna also assume lowercase, but here we're just duplicating the string
Screen Shot 2019-12-19 at 11 02 25 AM

src/AppActions.js
```diff
@@ -14,6 +14,12 @@ const setData = async (url, headers, options) => {
     return response.data;
 };
 
+const deleteData = async (url, headers, options) => {
```
AllenBW: ok, new plan, lets not add any actions for this one off table, they won't be reused anywhere else, instead take a look at the ViewHostAcks.js line 30, do something like that in the AddEditTopic modal (the `await API.delete(`${BASE_URL}/hostack/${host.id}/`);` thing)

ALSO, talked with @dkuc he reports two thingsssss
* updating a topic has to be done over PUT not post (and will include the slug in the url)
* what the mocks call `label` needs to be `tag`  (the one that doesn't autofill) and its gotta be passed when creating/updating as well

src/PresentationalComponents/TopicsAdminTable/TopicsAdminTable.js
```diff
@@ -1,8 +1,183 @@
...
+                }]
+            }]);
+        } else {
+            const rows = topicsArray.flatMap((value, key) => ([{
+                isOpen: false,
+                rule: value,
+                cells: [{
+                    title: <span key={key}> {value.name}</span>
+                }, {
+                    title: <span key={key}> {value.tags}</span>
```
AllenBW: this is `tag` (it changed a lil while ago)

src/AppActions.js
LI0131: Should the API calls be nested in a try-catch like they are in the ViewHostAcks modal?

LI0131:
@AllenBW In order to make sure that deletes aren't happening for topics that don't exist, should I make the `delete` button deactivated when a user presses `Create new topic` versus `Edit Topic`?

AllenBW:
OOOO i like that approach! disable it 👍🏽

AllenBW:
@LI0131 I made a tiny commit, it does this 
Screen Shot 2020-01-13 at 7 04 39 AM

Pleaseeeeee review, n if yer ok with it i think we can merge this 🥇 💝 🍫

LI0131:
@AllenBW That looks awesome! I'll go ahead and squash, and we can get this merged in!

PR #488, opened 2019-11-25
Ensure disable single system is true by default

AllenBW:
#### because this makes more sense, ya'd not be clickin on this less ya wanted to disable it for a single system 🤷‍♂ 
Screen Shot 2019-11-25 at 1 56 02 PM

plus brandon said this would be useful 🤗


AllenBW:
cc @skarekrow

skarekrow:
LGTM!

PR #489, opened 2019-11-26
Rule detail disabled notification is now a card (not a fake alert)

AllenBW:
as per https://marvelapp.com/i4441ie/screen/62092193

#### looks like
Screen Shot 2019-11-26 at 7 38 36 AM
Screen Shot 2019-11-26 at 7 37 59 AM


katierik:
Woot woot

"View Systems" should be sentence case (aka "View systems").

This is the 5 millionth time I've hit this, but the title style doesn't match what I'm getting from the design files. 
The 4th level [here ](https://www.patternfly.org/v4/documentation/react/components/text) seems to match the mocks, but isn't what I'm seeing above.
image

AllenBW:
@katierik we're using https://patternfly-react.surge.sh/patternfly-4/documentation/react/components/title#sizes xl title  as was indicated by the mocks

katierik:
It may be because of the font difference. Good for now

PR #498, opened 2019-12-09
Rules table, disable export when impacting = false 👨‍🏫

AllenBW:
includes https://github.com/RedHatInsights/frontend-components/pull/299
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-2883

#### tada
Screen Shot 2019-12-09 at 11 24 43 AM


bsquizz:
Cool... though it's not clear in your screenshot (doesn't look like a screenshot of disabled export) -- what will it look like when export is disabled? Is the button just grayed out? Does the user get any info on why it's grayed out via a mouse-over pop-up note or something like that?

AllenBW:
@bsquizz screen shot is of it disabled, both export buttons are grayed out, disabled icon dropdown's aren't a thing so what we can do is disable the action

katierik:
Eek. The export should *always* be available to the user, and never disabled. If the user has nothing selected, "export" should export out the entire list. 
QE is incorrect on this, export should *always* be enabled and working.

bsquizz:
That's not the use case here... The user has the filter selected to show rules which are not impacting systems.

There's nothing to export when that filter is turned on... Since the advisor export is intended to show you a list of rules that ARE impacting systems.

bsquizz:
Let me rephrase another way... If you select the filter to "show rules not impacting systems" and run an export, the end result is you get a list of all the systems impacted by all the rules. That's not good either.

The API is not capable of exporting simply a list of rules to you.

So we decided the best approach for now would be to disable export if that filter is selected.

I thought we all talked about this? 😆

katierik:
@bsquizz oh goodness we did I apparently have the memory of a goldfish. Here's what we had decided on according to my notes:
Temporary solution: disable the export button when the user hits one of these conditions
Permanent solution: rewrite API or design better error handling (more of the 1st to match consistency and the WYSIWYG-ness of the export)

dkuc:
@bsquizz @katierik 
Even more permanent solution: Export rules on the rules page and export hits on the hits page. Fitting of a square peg in a round hole goes away.

AllenBW:
@dkuc does the api support exporting hits? @katierik should this work pivot?

dkuc:
> @dkuc does the api support exporting hits? @katierik should this work pivot?

Yes.  It's the only thing we export currently.

Robcwilliams:
We do not have a "hits" page nor any plans to do so prior to summit - The purpose of the export functionality is to enable customers to get their results out of our product, not a list of rules. We need to find a solution in the current Information Architecture. 

Initial implementation of this was to export all - second pass is to provide granular views where export properly reflects the filters applied.

bsquizz:
> Initial implementation of this was to export all - second pass is to provide granular views where export properly reflects the filters applied.

Well... whether it was supposed to happen or not ... the second pass has actually been implemented for awhile at this point... it just has a couple glitches that Allen is sorting out here :laughing: 

The export button for the most part does export a list of rule hits based on what the rules table is filtered on. It's just that there's 2 filters a customer can select which would yield no rule hits (Showing "rules which are impacting no systems" and showing "rules which are disabled") ... so Allen's working on just disabling the export button if those filters are checked as the quick resolution until we refactor this whole thing.

AllenBW:
Okie dokie, given that this prevents the export of a list of rules (what export results become when you show ones without host hits) imma merge this controversial piece of 🌷 🌹 🌺

PR #500, opened 2019-12-10
Improve rule details small screen performance 

AllenBW:
####  finally usable at smaller size, looks the same at normal size
Screen Shot 2019-12-10 at 7 37 57 AM
Screen Shot 2019-12-10 at 7 37 40 AM


katierik:
Oh happy day!
I think the blue line is supposed to also be on the expanded pane, even in mobile.

AllenBW:
oooo it wasn't there previously, probably a pf-table bug, i'll look at it

AllenBW:
@katierik looks like a pf bug, here's a ss of the table expanded component 
Screen Shot 2019-12-10 at 8 09 59 AM

katierik:
Logged bug with PF

PR #505, opened 2019-12-12
Await setAck so the afterFn doesn't fire too soon

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-3747


AllenBW:
@Robcwilliams @skarekrow for catching this one,  💘 🤗

PR #528, opened 2020-01-14
Swaps topics cards for a topics table

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-3931

#### default view (the rest are still valid, but we renamed a column)
Screen Shot 2020-01-15 at 8 25 52 AM

#### sorting by affected, searching for rule
Screen Shot 2020-01-14 at 10 16 20 AM

#### when no topics exist
Screen Shot 2020-01-14 at 10 14 26 AM

#### sorting by affected
Screen Shot 2020-01-14 at 10 08 21 AM

#### no filter results
Screen Shot 2020-01-14 at 10 38 58 AM


terezanovotna:
Very nice job! 👍👍

Few improvements:

1. 


2.


3.



[Referencing pf](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state#main-content) and their empty state.

AllenBW:
1. its for the filterchips, no way we can kill if, would have to be done in fec
2. 👍 
3. 👍

AllenBW:
PS: still balanced 😁

AllenBW:
after talkin with @terezanovotna we:
* added the name column (width of two)
* added filtering by both name and description
* increased the width of search box (to accommodate lengthy helper text)

first ss in pr updated to reflect these changes

katierik:
> 1.the space is created by filterchips (a part of the fec primarytoolbar component), we could do away with filterchips here which would remove it, or the fix would have to happen in fec
> 2. 👍
> 3. 👍
> 
> Updated ss attached @terezanovotna
> Screen Shot 2020-01-14 at 1 56 03 PM
> Screen Shot 2020-01-14 at 1 54 31 PM

I'm surprised that we are stuck with the space from filter chips and can't show that only conditionally. We're stuck with this? Can platform update? Does PF need to update?

AllenBW:
@katierik no worries, am on it https://projects.engineering.redhat.com/browse/RHCLOUD-4002

its tough to notice it on all other tables because they use chips most of the time

PR #541, opened 2020-01-28
Fix default page when going from inv detail

karelhala:


AllenBW:
Tracking this, likely will close as the fix is located somewhere else 🤗

AllenBW:
closing cuz https://github.com/RedHatInsights/frontend-components/pull/335 is da fix

AllenBW:
woops, thats not the close button 😬

PR #544, opened 2020-01-30
Adds statefulness to Systems table

AllenBW:
tl;dr systems table now supports url params, maintains state when not in use, updates url params when table filters change, abstracts the similarities with rule table

fixes https://projects.engineering.redhat.com/browse/RHCLOUD-3973

### DIFF
--- a/src/PresentationalComponents/Common/Tables.js
+++ b/src/PresentationalComponents/Common/Tables.js
@@ -1,0 +1,22 @@
+// Builds returns url params from table filters, pushes to url if history object is passed
+export const urlBuilder = (filters, history) => {
+    const queryString = `?${Object.keys(filters).map(key => `${key}=${Array.isArray(filters[key]) ? filters[key].join() : filters[key]}`).join('&')}`;
+    history && history.replace({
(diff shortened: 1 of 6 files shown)


src/PresentationalComponents/Common/Tables.js
```diff
@@ -0,0 +1,28 @@
...
+    const queryString = `?${Object.keys(filters).map(key => `${key}=${Array.isArray(filters[key]) ? filters[key].join() : filters[key]}`).join('&')}`;
+    history && history.replace({
+        search: queryString
+    });
+    return queryString;
+};
+
+// transforms array of strings -> comma seperated strings, required by advisor api
+export const filterFetchBuilder = (filters) => (Object.assign({},
+    ...Object.entries(filters).map((filter) => {
```
Hyperkid123: You can use array destructing instead of declaring the constants :wink: 
```jsx
...Object.entries(filters).map(([filterName, filterValue]) => {
  ...
```

AllenBW: awesome catch!

PR #546, opened 2020-02-05
Updates deps to support fec changes

AllenBW:
in support of https://github.com/RedHatInsights/frontend-components/pull/340
similar to https://github.com/RedHatInsights/catalog-ui/pull/404
and https://github.com/RedHatInsights/sources-ui/pull/246

We're gonna call this part 1, doesn't completely close out https://projects.engineering.redhat.com/browse/RHCLOUD-4404 (cuz there are a few more pieces before da sizes get 🦐 sized) but its a large enough chunk that doesn't break anything that will save miserable rebases by merging it sooner


package.json
```diff
@@ -9,24 +9,24 @@
...
     "@patternfly/react-charts": "5.2.21",
-    "@patternfly/react-core": "3.134.2",
+    "@patternfly/react-core": "3.136.3",
     "@patternfly/react-icons": "3.14.39",
-    "@patternfly/react-table": "2.25.6",
+    "@patternfly/react-table": "2.26.5",
     "@patternfly/react-tokens": "2.7.25",
-    "@redhat-cloud-services/frontend-components": "0.0.47",
+    "@redhat-cloud-services/frontend-components": "0.0.47-beta",
     "@redhat-cloud-services/frontend-components-inventory-insights": "0.4.2",
```
Hyperkid123: You have to update all packages coming from `@redhat-cloud-services` if you want to get the full benefit of the bundle size optimizations. I recommend adding https://www.npmjs.com/package/webpack-bundle-analyzer to see some funky duplications and huge chunks in the app.

Basically, if you see anything coming from `@patternfly/react-core/dist/esm` or `@patternfly/react-icons/dist/esm`, then something is wrong. Either you forgot to update some packages, or I missed something in the PR. Everything from PF should come from `/dist/js`

karelhala:
As Martin said, you can give it a try with bundle analyzer and see how much this change removes unnecessary files.

package.json
AllenBW: had an issue importing the individual pf icons, as in, while are present, webpack config on @redhat-cloud-services/frontend-components-config doesn't appear to support it at the moment? 🤔

AllenBW: Screen Shot 2020-02-05 at 4 34 53 PM
so pretty, looks like getting those icons reduced is a good idea

Hyperkid123: Icons were one the main problem and were definetly resolved in the frontend components package. I will check out this pr a search in node modules from where its comming from.also you still have react-core esm modules inside the which is not right.

Hyperkid123: So I see that icons are coming from two packages: `inventory-insights` and `remediations`. Checking why are these packages importing the whole thing.

Hyperkid123: @AllenBW HA you forgot to update :skull: `@redhat-cloud-services/frontend-components-inventory-insights` and `@redhat-cloud-services/frontend-components-remediations`. I will test that now.

Hyperkid123: Ok so this is caused by the hot loaded inventory, specifically by these lines of code
```jsx
import * as reactCore from '@patternfly/react-core';
import * as reactIcons from '@patternfly/react-icons';
```

you are actually importing everything, and you have to di that in order to use the hot loaded inventory. This should be fixed in the component and it should provide a clear list of components to be used. @karelhala I think you have homework :smile: 

But you should still update the packages from the comment above.

Hyperkid123: Also, you did use absolute import paths for the react-icons. So I guess there is not much we can do until `loadInventory` is fixed. I will check the list.

AllenBW: I THOUGHT THAT WAS WHY!!! in making everything smaller, thought... hmm, but we're still grabbing everything for inventory... was my final thought before sleep, to see if inventory would hollar if we took away... everything...  gonna update other packages! also did try absolute path for icons, get a bunch of "react components can't be null" errors on it

Hyperkid123: Icons use default exports not named as react-core, that is probably why. I tried it with your PR and it works.

AllenBW: well hot darn, i'll check that out n get an update to this pr up soon! <3

Hyperkid123: Yup. I think we've found the issue. There are old versions of frontend component packages in frontend components :D

AllenBW: Screen Shot 2020-02-06 at 8 15 38 AM


well its notta big change, but a lil one... N BESIDES once we figure out inventory stuff it'll be much bigger!

Hyperkid123: Yup as I said there are older deps in the frontend components. I am close to a solution right now. Will send an update once its fixed

Hyperkid123:
@AllenBW So I might have a fix but thanks to the trash implementation of workspaces in Lerna :cry: I can't test it locally until its actually released which will be on Friday. I will test it on Monday since I am on PTO tomorrow

package.json
AllenBW: you ok if I merge this pr for the time being (cuz it touches a LOTTA FILES) n revisit when yah update?

Hyperkid123: I am ok with that.

PR #553, opened 2020-02-13
Bottom pagination was missing onPerpageSelect

AllenBW:
something probably changed somewhere and we didn't change with it


skarekrow:
🥇

PR #567, opened 2020-03-03
Release branches to support master-beta -> master -> master-stable

AllenBW:
in order to support the new dashboard n app consolidation efforts, advisor has gotta be able to target qa/ci beta specifically, this app allows us to do that while keeping everything else the same

### DIFF
--- a/.travis/custom_release.sh
+++ b/.travis/custom_release.sh
@@ -2,33 +2,23 @@
 set -e
 set -x
 
-if [ "${TRAVIS_BRANCH}" = "master" ]; then
-    # for a limited time prod too
-    for env in ci qa prod
-    do
-        echo
+if [ "${TRAVIS_BRANCH}" = "master-beta" ]; then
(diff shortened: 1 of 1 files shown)


.travis/custom_release.sh
```diff
@@ -2,33 +2,24 @@
...
+elif [ "${TRAVIS_BRANCH}" = "master" ]; then
+    echo
+    echo "PUSHING prod-beta"
+    rm -rf ./dist/.git
+    .travis/release.sh "prod-beta"
+elif [ "${TRAVIS_BRANCH}" = "master-stable" ]; then
+    for env in ci qa prod; do
+        
         echo
         echo "PUSHING ${env}-beta"
```
skarekrow:
```suggestion
        echo "PUSHING ${env}-stable"
```

AllenBW: AWESOME CATCH @skarekrow <3

PR #597, opened 2020-03-24
[RFR] Add large margins to Affected Systems Title

LI0131:
This PR is in response to RHCLOUD-5244 (page 3 of the google doc). We are adding large bottom margins to the `Affected Systems` title. 
Screen Shot 2020-03-24 at 11 01 25 AM

### DIFF
--- a/src/SmartComponents/Rules/Details.scss
+++ b/src/SmartComponents/Rules/Details.scss
@@ -6,6 +6,10 @@
   margin-bottom: var(--pf-global--spacer--md);
 }
 
+.titleOverride {
+  margin-bottom: var(--pf-global--spacer--lg);
+}
+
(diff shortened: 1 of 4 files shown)


AllenBW:
Gonna also have to do this to topic details view as well
Screen Shot 2020-03-24 at 3 18 20 PM

src/SmartComponents/Rules/Details.scss
```diff
@@ -9,3 +9,7 @@
 .pf-c-card__body {
   padding-bottom: var(--pf-global--spacer--md);
 }
+
+.pf-c-title {
```
AllenBW: this is already defined globally, does the class name thing above fixit?

PR #613, opened 2020-04-08
Use inventory when showing list of systems

karelhala:
animation

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -1,66 +1,128 @@
 /* eslint camelcase: 0 */
 import * as AppActions from '../../AppActions';
+import * as pfReactTable from '@patternfly/react-table';
+import * as reactRouterDom from 'react-router-dom';
 
-import { Pagination, PaginationVariant } from '@patternfly/react-core/dist/js/components/Pagination/Pagination';
-import React, { useEffect, useState } from 'react';
-import { Table, TableBody, TableHeader, cellWidth, sortable } from '@patternfly/react-table';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import { connect, useStore } from 'react-redux';
 import { paramParser, urlBuilder } from '../Common/Tables';
 
-import { DateFormat } from '@redhat-cloud-services/frontend-components/components/DateFormat';
 import Failed from '../Loading/Failed';
-import { Link } from 'react-router-dom';
 import Loading from '../Loading/Loading';
-import MessageState from '../MessageState/MessageState';
-import { PrimaryToolbar } from '@redhat-cloud-services/frontend-components/components/PrimaryToolbar';
 import PropTypes from 'prop-types';
-import SearchIcon from '@patternfly/react-icons/dist/js/icons/search-icon';
-import { TableToolbar } from '@redhat-cloud-services/frontend-components/components/TableToolbar';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications';
-import { connect } from 'react-redux';
 import debounce from '../../Utilities/Debounce';
+import { getRegistry } from '@redhat-cloud-services/frontend-components-utilities/files/Registry';
 import { injectIntl } from 'react-intl';
 import messages from '../../Messages';
 import routerParams from '@redhat-cloud-services/frontend-components-utilities/files/RouterParams';
+import { systemReducer } from '../../AppReducer';
 
 const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters, setFilters, history, selectedTags }) => {
-    const [cols] = useState([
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -42,16 +47,31 @@ const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters
...
+                pfReactTable
+            });
+            getRegistry().register({
+                ...mergeWithEntities(systemReducer(cols, INVENTORY_ACTION_TYPES))
+            });
+
+            const { InventoryTable } = inventoryConnector(store);
+            setInventory(() => InventoryTable);
+        })();
+    }, []);
```
AllenBW, line 74: should this have the dependencies`[cols, store]`  ?

src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -86,10 +106,14 @@ const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters
     }, [filters, history]);
 
     useEffect(() => {
-        if (!filterBuilding && filters.sort !== undefined) {
```
AllenBW, line 89: need filterbuilding here, otherwise on page refresh you'll pound the api 4 times (we read sort params from the url if they're present and set them, if not present we'll push the default which is updated)

src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -158,38 +144,35 @@ const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters
...
-            filterConfig={{ items: filterConfigItems }}
-            activeFiltersConfig={activeFiltersConfig}
-        />
-        {systemsFetchStatus === 'fulfilled' &&
-            <Table aria-label={'rule-table'} sortBy={sortBy} onSort={onSort} cells={cols} rows={rows}>
-                <TableHeader />
-                <TableBody />
-            </Table>
+        {
+            InventoryTable && systemsFetchStatus !== 'failed' ?
```
AllenBW, line 157: Wondering if we shouldn't leave the loading state for the bridge landing on the page and loading the inventory table, presently s just kinda white space 🤔

src/PresentationalComponents/SystemsTable/SystemsTable.js
karelhala: When I left this on the initial filter was not applied. Meaning when you refreshed page the URL was changed, but table was not sorted. I'll give that a try and see if it'll actually fire multiple requests.

src/PresentationalComponents/SystemsTable/SystemsTable.js
karelhala: Not really, this needs to happen only on initial render.

src/PresentationalComponents/SystemsTable/SystemsTable.js
karelhala: Yeah, good point! Updated.

src/PresentationalComponents/SystemsTable/SystemsTable.js
AllenBW: ooo in that case would yah mind throwing ah  ` // eslint-disable-next-line react-hooks/exhaustive-deps` in the line above?

karelhala: Added, thanks for pointing it out!

AllenBW:
Looks like offset isn't paying attention to url (on refresh defaults to 0 despite there being a value) workin on fixing that atm

src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -1,66 +1,93 @@
...
     useEffect(() => {
         filters.display_name === undefined ? setSearchText('') : setSearchText(filters.display_name);
     }, [filters.display_name]);
 
     useEffect(() => {
-        const filter = { ...filters };
-        const text = searchText.length ? { display_name: searchText } : {};
-        delete filter.display_name;
-        setFilters({ ...filter, ...text });
+        if (Boolean(searchText) !== Boolean(filters.display_name)) {
```
AllenBW, line 85: for posterity, this breaks url update, cuz we're not removing the search field from the filter object, (that's what builds the url)

AllenBW:
This fixes https://projects.engineering.redhat.com/browse/RHCLOUD-6060

AllenBW:
@karelhala worked this ah bit to ensure no regressions, yer 👀 would be greatly appreciated!

AllenBW:
#### Updated gif showin whats what

sys-inv-table

karelhala:
@AllenBW Looking hella good! I am the original author, so I can't ACK it. But I'd def do it!

PR #624, opened 2020-04-16
Prevent multiple rendering when changing tabs

karelhala:
This will render the button on initial render and everytime loading changes, so it'll prevent multiple renders.


AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-588 4️⃣ real

PR #635, opened 2020-04-28
Adds sorting to affected systems table

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-5679

need to squash this b, will do when merging, also need to switch sorting from last_seen to updated, but that capability doesn't exist yet

This needs api support (add sorting (by updated, display_name, -updated, -display_name) to /rule/{rule_id}/systems/ ) before it can become... 

what it is destine to become!


AllenBW:
closing in favor of #654 AND  #655

PR #640, opened 2020-05-05
Tags query param mirrors component selectedTag value 💪😝🥺😵

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-6366

* as tags are selected, they'll show up in the url or be removed from it
* when navigating to ah page that has tags in the param, values on the page will reflect those tags
* navigating between pages, they'll be preserved, tags
* most importantly, everything works with even doofy tags `insights-client/contact=jcordes%40redhat.com` ie `tags=insights-client%2Fcontact%3Djcordes%2540redhat.com`

#### works from both the dropdown and the modal
Screen Shot 2020-05-05 at 1 42 14 PM

### DIFF
--- a/src/PresentationalComponents/TagsToolbar/TagsToolbar.js
+++ b/src/PresentationalComponents/TagsToolbar/TagsToolbar.js
@@ -12,11 +12,14 @@ import PropTypes from 'prop-types';
 import TagIcon from '@patternfly/react-icons/dist/js/icons/tag-icon';
 import { addNotification } from '@redhat-cloud-services/frontend-components-notifications';
 import { connect } from 'react-redux';
-import { injectIntl } from 'react-intl';
+import intersection from 'lodash/intersection';
 import messages from '../../Messages';
 import { setSelectedTags } from '../../AppActions';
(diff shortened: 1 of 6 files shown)


src/PresentationalComponents/TagsToolbar/TagsToolbar.js
```diff
@@ -12,29 +12,47 @@ import PropTypes from 'prop-types';
...
             }
         };
 
         fetchTags();
-    }, [intl]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []);
+
+    useEffect(() => {
+        console.error(window.location.search, selectedTags);
```
karelhala, line 52: Do you want to console.error these values?

AllenBW: OOOO GOOD POINT

AllenBW:
and awaaaaaayyyyyyyyy we go! gonna hold off on prod-stable till @skarekrow says :godmode:

PR #644, opened 2020-05-13
Add TagToolbar component to Topics detail page

mhuth:
RHCLOUD-6265


mhuth:
After shots of adding the host tags toolbar to the topics detail page:
hosttags_filter_for_topic_details1
hosttags_filter_for_topic_details2

PR #645, opened 2020-05-15
Adds support for local storage, persists state with valid value

AllenBW:
fixes https://projects.engineering.redhat.com/browse/RHCLOUD-6403

SO what we do here with this work is propagate advisor store into sessionStorage and use it to hydrate the store on app load. The goal being, when you navigate away from advisor then return all filter settings will be preserved/restored.

### DIFF
--- a/src/AppReducer.js
+++ b/src/AppReducer.js
@@ -36,7 +36,8 @@ const initialState = Immutable({
     selectedTags: null
 });
 
-export const AdvisorStore = (state = initialState, action) => {
+// eslint-disable-next-line new-cap
+export const getAdvisorStore = (previousState) => (state =  Immutable(previousState) || initialState, action) => {
     switch (action.type) {
         case `${ActionTypes.RULE_FETCH}_PENDING`:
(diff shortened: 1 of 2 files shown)


src/Store/LocalStorage.js
```diff
@@ -0,0 +1,32 @@
...
+
+        if (previousAdvisorStore && !AdvisorStore.sessionChecked) {
+            previousAdvisorStore.sessionChecked = true;
+            setStore(previousAdvisorStore);
+        } else {
+            sessionStorage.setItem('AdvisorStore', JSON.stringify(AdvisorStore));
+        }
+    }, [AdvisorStore, setStore]);
+
+    return <React.Fragment />;
```
rvsia, line 21: You can return `null`

AllenBW:
@rvsia thanks for the 👀 ! Will take ah look 🙇

AllenBW:
@rvsia hella awesome feedback a *FAR* more elegant solution ❤️  🍍 🙇

PR #649, opened 2020-05-20
Remove SMBloris rule from cveToRuleid mapping as its not a CVE rule

mhuth:


mhuth:
FFS, travis is being a pain

PR #650, opened 2020-05-20
Update cveToRuleid mapping to remove SMBloris rules

mhuth:
SMBloris isn't actually a CVE rule afterall, so remove it from the map in cveToRuleid.js

### DIFF
--- a/src/cveToRuleid.js
+++ b/src/cveToRuleid.js
@@ -423,22 +423,6 @@ export const cveToRuleid = [
         ]
     },
     {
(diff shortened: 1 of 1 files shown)


mhuth:
Note, I stuffed up an earlier PR, closed it and deleted the branch quickly, and travis wasn't happy about that.  When I tried to submit a new PR, travis seemed to think it was related to the old PR and kept failing the tests.  I had to change the commit id and branch name and travis seems happy(er) now.

PR #663, opened 2020-06-08
Use radio buttons instead of checkboxes for some RuleTable filters

mhuth:
Seems to make more sense to use radio buttons for these filters.  They are either true or false, seems a bit silly to be able to select both.


mhuth:
radio_buttons

mhuth:
Putting this on hold for the time being.  Unfortunately the radio buttons in the UI retain their selection even after clearing the filters.  Note that the filter itself is cleared, but the UI maintains their selection until you move away from the radio buttons and then come back to them.  Then they are cleared in the UI.  Is it possible to somehow clear the 'selected' attribute from the radio button elements in the UI when clearing the filters in the React component??  That should also cause the UI elements to be cleared as well.

AllenBW:
gonna close this for now, feel free to reopen new pr when/if we go this route in the future!

PR #670, opened 2020-06-11
Filter systems table by whether or not the columns have hits

mhuth:
fixes: https://projects.engineering.redhat.com/browse/RHCLOUD-6434

### DIFF
--- a/src/AppConstants.js
+++ b/src/AppConstants.js
@@ -160,3 +160,12 @@ export const FILTER_CATEGORIES = {
         ]
     }
 };
+
+export const SYSTEM_FILTER_CATEGORIES = {
+    hits: {
+        type: 'checkbox', title: 'Total Risk', urlParam: 'hits', values: [
+            { label: 'All systems', text: 'All systems', value: 'all' },
+            ...FILTER_CATEGORIES.total_risk.values
+        ]
+    }
+};
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -5,7 +5,7 @@ import * as reactRouterDom from 'react-router-dom';
 
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { connect, useStore } from 'react-redux';
-import { paramParser, urlBuilder } from '../Common/Tables';
+import { paramParser, urlBuilder, filterFetchBuilder, pruneFilters } from '../Common/Tables';
 
 import Failed from '../Loading/Failed';
 import Loading from '../Loading/Loading';
@@ -18,6 +18,7 @@ import { injectIntl } from 'react-intl';
 import messages from '../../Messages';
 import routerParams from '@redhat-cloud-services/frontend-components-utilities/files/RouterParams';
 import { systemReducer } from '../../AppReducer';
+import { SYSTEM_FILTER_CATEGORIES as SFC } from '../../AppConstants';
 
 const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters, setFilters, selectedTags }) => {
     const inventory = useRef(null);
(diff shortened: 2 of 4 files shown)


mhuth:
No filter, so only display systems with hits ...
no_filter_only_systems_with_hits

mhuth:
Choosing All Systems causes all systems (including those without hits) to be displayed  ...

all_systems_includes_systems_with_no_hits

mhuth:
Choosing Low will display only systems with Low Risk hits.  The systems might have other risk hits as well, but systems without Low Risk hits will be filtered out.  Similar will happen when selecting the other types of Risk hits.  Note, if multiple are selected, then it does a logical 'AND'.  Eg if both Moderate and Low are selected, then only systems with both Moderate risk hits and Low risk hits will be displayed.

low_risk_filter_only_systems_with_low_risk_hits

mhuth:
Gonna make some changes.  I'll have the systems table always supply a hits= parameter to the API to control what systems are displayed or not

terezanovotna:
looks good! thanks for finding the correct mockup for this!

src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -44,22 +45,84 @@ const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters
...
+        }
     }];
 
+    const buildFilterChips = () => {
+        const localFilters = { ...filters };
+        delete localFilters.sort;
+        delete localFilters.offset;
+        delete localFilters.limit;
+
+        const prunedFilters = Object.entries(localFilters);
```
AllenBW, line 91: aside from the deletes in the beginning, this function looks identical to the rec table filter builder, you know what that means 🙃 

abstract it to common! (specifically lines 91 to 106) great news is, since its working presently, breaking it to abstract shouldn't be too painful!

src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -144,7 +207,7 @@ const SystemsTable = ({ systemsFetchStatus, fetchSystems, systems, intl, filters
             paramsObject.offset === undefined ? paramsObject.offset = 0 : paramsObject.offset = Number(paramsObject.offset[0]);
             paramsObject.limit === undefined ? paramsObject.limit = 10 : paramsObject.limit = Number(paramsObject.limit[0]);
             setFilters({ ...filters, ...paramsObject });
-        } else if (filters.limit === undefined || filters.offest === undefined) {
+        } else if (filters.limit === undefined || filters.offset === undefined) {
```
AllenBW, line 210: noice 👀 🧶 😍 🤩

src/AppConstants.js
```diff
@@ -158,5 +158,14 @@ export const FILTER_CATEGORIES = {
             { label: intlHelper(intl.formatMessage(messages.enabled), intlSettings), value: 'true' },
             { label: intlHelper(intl.formatMessage(messages.disabled), intlSettings), value: 'false' }
         ]
+    },
+    hits: {
```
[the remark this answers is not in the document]
mhuth: Done!

src/PresentationalComponents/SystemsTable/SystemsTable.js
mhuth: Done!

src/PresentationalComponents/SystemsTable/SystemsTable.js
mhuth: :bow:

mhuth:
Ok, made a few changes.  Made the changes as suggested by @AllenBW   There is now a required hits= parameter to be passed to the API to make this work.  By default it passes 'hits=yes' if no filters are set.  Otherwise it passes an array of values corresponding to what is selected by the user. 

Also, changed the logic so that if multiple filters are selected, it does a logical 'OR' now. Eg if both Moderate and Low are selected, then systems with either Moderate risk hits OR Low risk hits will be displayed.

src/AppConstants.js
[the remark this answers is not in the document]
AllenBW: not quite, that `...FILTER_CATEGORIES.total_risk.values` takes the above values from the other object n adds em to our new constant... which essentially the same, cept that "all systems" jazz in the front

mhuth:
Please test this well!

src/AppConstants.js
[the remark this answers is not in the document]
mhuth: Makes sense.  Done!

PR #672, opened 2020-06-15
Add Filtering to Tags Selector

LI0131:
This PR is in response to https://projects.engineering.redhat.com/browse/RHCLOUD-6739, which requests the implementation of filtering for the tags selection tool.

Before:
Screen Shot 2020-06-15 at 5 05 39 PM

After:
Screen Shot 2020-06-15 at 5 05 59 PM

### DIFF
--- a/src/PresentationalComponents/TagsToolbar/TagsToolbar.js
+++ b/src/PresentationalComponents/TagsToolbar/TagsToolbar.js
@@ -7,11 +7,17 @@ import { Tooltip, TooltipPosition } from '@patternfly/react-core/dist/js/compone
 import API from '../../Utilities/Api';
 import { BASE_URL } from '../../AppConstants';
 import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
+import { DEBOUNCE_DELAY } from '../../AppConstants';
+import { Divider } from '@patternfly/react-core/dist/js/components/Divider/Divider';
+import { InputGroup } from '@patternfly/react-core/dist/js/components/InputGroup/InputGroup';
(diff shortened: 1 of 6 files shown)


src/PresentationalComponents/TagsToolbar/TagsToolbar.js
```diff
@@ -1,6 +1,8 @@
 import './_TagsToolbar.scss';
 
+import { Divider, InputGroup, TextInput } from '@patternfly/react-core';
```
AllenBW, line 3: checkout how we bring in the select component below, for the sake of small bundles, this needs to be dedicated imports for each piece

this also applies to line 5, the icons, lookin around the codebase yah should find examples of how this is done

src/PresentationalComponents/TagsToolbar/TagsToolbar.js
```diff
@@ -88,6 +99,21 @@ const TagsToolbar = ({ selectedTags, setSelectedTags }) => {
             ariaLabelledBy='select-group-input'
             isDisabled={tags.length === 0}
         >
+            <InputGroup className='tags-filter-group'>
+                <TextInput
+                    placeholder='Filter by tags'
```
AllenBW, line 104: lets add this to messages.js

AllenBW:
@LI0131 jumped in! three things to tweak...

-  add debounce to the input
- when searching on an empty field, lets return the original list
- looks like search is case sensitive, thinking as a kindness it oughta be case insensitive 

those aside, LOOKIN HECKAH GOOOOODDDD 🤗

PR #678, opened 2020-06-22
More gracefully handle incorrect filter value

mhuth:
Found a small bug after adding the total_risk filter.  It can reproduced by manually entering a value for 'hits= ' other than the expected values of all, yes, 1, 2, 3, 4.  Another value the API takes is 'no' but if I manually use that then I get a blank section of screen instead of the systems table, and the error in the attached screenshot.  My fix at least prevents the error and the systems table is rendered properly, but perhaps there's a better way of handling this problem than what I have used?  

systems_table_bug


AllenBW:
OOOOO good catch... solving it this issue in the filter chips builder s a lil too late in the process tho... yields this result
Screen Shot 2020-06-22 at 8 58 15 AM

so the problem shows when keying in invalid url, maybe we oughta have additional sanitizing of that input before accepting it as valid filters 🤔

AllenBW:
PS: AWESOME FIND @mhuth

mhuth:
@AllenBW Ok, made a slight change and the value now shows up in the filter chip.  Not sure why I didn't do this originally.

systems_table_bug2

PR #681, opened 2020-06-25
Add CVE_2020_14298_docker|CVE_2020_14298_DOCKER rule to cveToRuleid.js

mhuth:


mhuth:
@AllenBW Get thee to prod!

AllenBW:
closin' time... you don't have to go home but you can't stay here...

mhuth:
Ooops, seems this wasn't necessary afterall.  As Advisor doesn't handle CVE's anymore, this change isn't necessary.  Apparently the shared engine is the place where CVEs are handled now.  Sorry for the false emergency.

PR #689, opened 2020-07-06
Make dropdowns text only

LI0131:
This PR removes the battery and shield icons from the dropdown menus used for filtering.

Before:
Screen Shot 2020-07-06 at 10 10 35 AM

After:
Screen Shot 2020-07-06 at 10 09 42 AM

Screen Shot 2020-07-06 at 10 09 33 AM


LI0131:
This change will also need to be mirrored in https://github.com/RedHatInsights/frontend-components/blob/master/packages/inventory-insights/src/Insights.js

PR #691, opened 2020-07-14
Add key=value to tags toolbar

mhuth:
The inventory tags dropdown has key=value.  Modifying ours to match.  Also allowing the width of the dropdown to change to match the max width of the longest key=value text.


mhuth:
Showing key=value for each selection and the wider dropdown list...

widen_tags_toolbar

mhuth:
> M good with displaying the key-value, but the growing dropdown is a no-go, some of those things get hella long, n would be super icky to look at, i thought we had the tooltip for things that overflow the dropdown ?

Yeah we do, but the text in the dropdown didn't look quite right.  I can set the width to about 400px.  That seemed to be a bit better.

AllenBW:
@mhuth m good with 400! but its gotta be fixed, between surprise super long tags and whatever the results the search input yields, anything not fixed would be all over the place 🤗

mhuth:
@AllenBW try that.  Can't say my CSS is too good either, but this seems to work ok, and the `text-overflow: ellipsis` works too.  Ultimately unimportant stuff, but what the heck

PR #694, opened 2020-07-16
Update grading system

LI0131:
This PR is in response to RHCLOUD-7516 (not sure why the branch is RHCLOUD-6757 🤷‍♂️). It updates the current grading system to a lovely new system which is detailed in the mocks which can be found -> https://marvelapp.com/prototype/8d7cj33/screen/70993112. Here's what the new system looks like:

Screen Shot 2020-07-16 at 12 21 26 PM


LI0131:
One thing to note is that the current implementation of the [SeverityLine](https://github.com/RedHatInsights/frontend-components/blob/master/packages/charts/src/Charts/SeverityLine/SeverityLine.js) component does not support truncation of intl strings. So we are not able to truncate strings which exceed the maximum width defined for a title. The results is wrapping strings.

Screen Shot 2020-07-16 at 12 26 29 PM

src/AppConstants.js
```diff
@@ -62,67 +62,69 @@ export const SEVERITY_MAP = {
     'medium-risk': 2,
     'low-risk': 1
 };
+// intlHelper seems to squash the use of HTML rich-text formatting
```
AllenBW, line 65: naop theres a way to do this 🤤  checkout `rulesDetailsTotalriskBody || overviewActionCallNoSystemsBody || overviewDeployBody ||` in messages.js

🦖 🎉

LI0131: Oooo I'll check this out 🎉 :godmode:

LI0131: So looked at those messages... and _think_ the messages below are formatted the same way. It seems like the `intlProvider` behind `intlHelper` doesn't treat those HTML strings the same way as the `intlProvider` we are using in Advisor. (e.g. it will load as something like `<strong>low</strong>` rather than **low** 😬 )

LI0131:
This also updates the `/recommendations/:id` page

Screen Shot 2020-07-16 at 4 56 21 PM

terezanovotna:
For the last recommendation page, 

- there seems to be switched grading (Total risk - high), but when I look at Total risk variables - they are: (Critical, Important, Moderate, Low). What happened here?
- the labels should have rounded corners

For reference - this is what we have:
Total risk (Critical, Important, Moderate, Low)
Risk of change (High, Moderate, Low, Very low)

src/AppConstants.js
AllenBW: mmmmm so if intlHelper wont do the formatting, lets switch to the other one, gotta keep things translated

LI0131:
@terezanovotna Looks like the mocks were a showing the Risk of Change labels slightly differently. I went ahead and reverted back to the labels you suggested 😁 I'm also pushing an update to fec to fix the Total Risk labels, but looking into the rounded corners before then 👍

package.json
```diff
@@ -48,8 +48,9 @@
     "@patternfly/react-table": "4.8.6",
     "@patternfly/react-tokens": "4.5.2",
     "@react-pdf/renderer": "1.6.9",
-    "@redhat-cloud-services/frontend-components": "2.1.3",
-    "@redhat-cloud-services/frontend-components-inventory-insights": "2.2.0",
+    "@redhat-cloud-services/frontend-components": "^2.3.1",
```
AllenBW, line 51: gotta kill da carrot, we lock down on specific versons

src/AppConstants.js
AllenBW: @LI0131 gotta switch these over 😬 🤗

src/PresentationalComponents/RuleDetails/RuleDetails.js
```diff
@@ -93,48 +84,105 @@ const RuleDetails = ({ children, rule, intl, topics, header, isDetailsPage }) =>
...
-                                    likelihood: AppConstants.LIKELIHOOD_LABEL[rule.likelihood] || intl.formatMessage(messages.undefined),
-                                    impact: AppConstants.IMPACT_LABEL[rule.impact.impact] || intl.formatMessage(messages.undefined),
-                                    strong(str) { return <strong>{str}</strong>; }
-                                })}</Text>
-                            </TextContent>
+                            <Split hasGutter>
+                                <SplitItem>
+                                    <InsightsLabel value={rule.total_risk}/>
+                                </SplitItem>
+                                <SplitItem isFilled/>
```
AllenBW, line 91: hasGutter introduces enough space here... less this is specifically requested, don't think we need this

src/PresentationalComponents/RuleDetails/RuleDetails.js
```diff
@@ -93,48 +84,105 @@ const RuleDetails = ({ children, rule, intl, topics, header, isDetailsPage }) =>
...
-                                <Text component={TextVariants.small}>
-                                    {AppConstants.RISK_OF_CHANGE_DESC[resolutionRisk]}
-                                </Text>
-                            </TextContent>
+                        <StackItem className={`pf-u-display-inline-flex alignCenterOverride pf-u-pb-sm pf-u-pt-sm`}>
+                            <Split hasGutter>
+                                <SplitItem>
+                                    <InsightsLabel text={AppConstants.RISK_OF_CHANGE_LABEL[resolutionRisk]} value={resolutionRisk} hideIcon/>
+                                </SplitItem>
+                                <SplitItem isFilled/>
```
AllenBW, line 137: again, not sure we need the extra padding

src/PresentationalComponents/RuleDetails/RuleDetails.js
```diff
@@ -93,48 +84,105 @@ const RuleDetails = ({ children, rule, intl, topics, header, isDetailsPage }) =>
...
+                                                        intl.formatMessage(
+                                                            AppConstants.RISK_OF_CHANGE_DESC[resolutionRisk], {
+                                                                strong(str) { return <strong>{str}</strong>; }
+                                                            }) : intl.formatMessage(messages.undefined)}
+                                                </Text>
+                                            </TextContent>
+                                        </StackItem>
+                                        <StackItem>
+                                            {rule.reboot_required ? <Split>
+                                                <SplitItem>
```
AllenBW, line 153: split necessary here? think these two can just be placed beside each other in a span? (same for the one below)

AllenBW:
ah few more thoughts...

Screen Shot 2020-07-17 at 1 42 17 PM
tooltip text rendering s a lil borked

Screen Shot 2020-07-17 at 1 46 44 PM
thinkin we're gonna need to give ansible column more room else it looks borked also

Screen Shot 2020-07-17 at 1 49 40 PM
this text is a lil grayer rather than black likely cuz yer using the <Text> component, mock shows it as black(ish)

src/PresentationalComponents/RuleDetails/RuleDetails.js
LI0131: So for this and the below, the mocks seem to want the right side of the top and bottom sections to be aligned. If I take out the extra `<SplitItem isFilled/>` they will appear at different distances. If that's alright then I can take it out 😁

src/PresentationalComponents/RuleDetails/RuleDetails.js
LI0131: same as above

package.json
LI0131: 🔐

src/AppConstants.js
LI0131: Wait what did you previous comment mean? I think I misunderstood. Do we want to try to use the `intlProvider` from advisor? That's kind of what I'm doing now, by moving translation into the RuleDetails component. Butttt there could be a better way of doing it

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -41,13 +41,14 @@ import routerParams from '@redhat-cloud-services/frontend-components-utilities/f
...
+        { title: intl.formatMessage(messages.name), transforms: [sortable, cellWidth(30)] },
         { title: intl.formatMessage(messages.added), transforms: [sortable, cellWidth(15)] },
         { title: intl.formatMessage(messages.totalRisk), transforms: [sortable, cellWidth(15)] },
-        { title: intl.formatMessage(messages.systems), transforms: [sortable] },
+        { title: intl.formatMessage(messages.riskofchange), transforms: [sortable, cellWidth(15)] },
+        { title: intl.formatMessage(messages.systems), transforms: [sortable, cellWidth(15)] },
         {
             title: <React.Fragment><AnsibeTowerIcon size='md' /> {intl.formatMessage(messages.ansible)}</React.Fragment>,
-            transforms: [sortable],
+            transforms: [sortable, cellWidth(15), fitContent],
```
LI0131, line 51: @AllenBW `fitContent` should fix the tooltip issue

LI0131:
Here's where we are at now:

Screen Shot 2020-07-17 at 3 26 21 PM

src/AppConstants.js
LI0131: Had to import `React` for the HTML to work 💀 😩

src/AppConstants.js
```diff
@@ -63,23 +64,43 @@ export const SEVERITY_MAP = {
     'low-risk': 1
 };
 export const RISK_OF_CHANGE_DESC = {
-    1: intlHelper(intl.formatMessage(messages.riskOfChangeTextOne), intlSettings),
-    2: intlHelper(intl.formatMessage(messages.riskOfChangeTextTwo), intlSettings),
-    3: intlHelper(intl.formatMessage(messages.riskOfChangeTextThree), intlSettings),
-    4: intlHelper(intl.formatMessage(messages.riskOfChangeTextFour), intlSettings)
+    1: intlHelper(intl.formatMessage(
+        messages.riskOfChangeTextOne, { strong(str) { return <strong>{str}</strong>; } }
```
AllenBW, line 68: _ducking_ cant help but wonder if the `strong(str) { return <strong>{str}</strong>;` could be declared once, used everywhere 🤔  (not blocking merge of this on that or anything... just 🤔 )

locales/translations.json
```diff
@@ -148,6 +156,8 @@
   "dateDisabled": "Date disabled",
   "hostAckModalTitle": "Recommendation has been disabled for:",
   "systemName": "System name",
+  "noSystemReboot": "System reboot is <strong> not </strong> required",
+  "systemReboot": "System reboot <strong> is </strong> required",
```
AllenBW, line 160: looks like this could be the case where yah have this string once, and pass in the `is` / `is not` 🤔 BUT ALSO period at the end of this

src/Messages.js
```diff
@@ -295,9 +295,8 @@ export default defineMessages({
     rulesDetailsTotalriskBody: {
         id: 'rulesdetails.totalriskbody',
         description: 'Text explaining the total risk value of this recommendation',
-        defaultMessage: `The <strong>likelihood</strong> that this will be a problem is
-        {likelihood}. The <strong>impact</strong> of the problem would be
-        {impact} if it occurred.`
+        defaultMessage: `The total risk of this remediation is <strong>{risk}</strong>,
```
AllenBW, line 298: period at the end of this 🤔

locales/data.json
```diff
@@ -51,10 +51,14 @@
...
     "filterTagsInToolbar": "Filter by tags",
+    "filterTagsInModal": "Filter tags",
+    "tagsModalName": "Name",
+    "tagsModalValue": "Value",
+    "tagsModalSources": "Tag Sources",
     "allSystems": "All systems",
     "knowledgebasearticle": "Knowledgebase article",
     "riskofchange": "Risk of change",
-    "rulesdetails.totalriskbody": "The <strong>likelihood</strong> that this will be a problem is\n        {likelihood}. The <strong>impact</strong> of the problem would be\n        {impact} if it occurred.",
+    "rulesdetails.totalriskbody": "The total risk of this remediation is <strong>{risk}</strong>,\n                        based on the combination of likelihood and impact to remediate",
```
AllenBW, line 61: think a period on the end of this one is necessary 🙃👩‍🏭

src/AppConstants.js
LI0131: I like that idea 🤔 would something like that belong as a new file in the `Utilities` directory?

AllenBW: OOO yeah! in utilities could work... make it like an intl helper file or something, we could kill ah lotta repetition that way... cuz <strong> s used in quite a few places

PR #695, opened 2020-07-17
Test for using Advisor's /systems_detail endpoint for rule system list

mhuth:
This is just a drill.  Repeat.   This is just a drill.


mhuth:
Nah, this isn't going anywhere for a while.  I'll close it and open another when I have something worth reviewing.

PR #701, opened 2020-07-21
Workaround for patternfly checkForValue.toString error in Select comp…

mhuth:
Perhaps we can use this in the meantime, until a more permanent solution is forthcoming.


mhuth:
Potential fix to patternfly: https://github.com/patternfly/patternfly-react/compare/master...mhuth:fix_undefined_valueToCheck

... but gotta jump through all the 'contributing patches' hoops first.

mhuth:
FWIW, upstream patternfly issue and PR:
https://github.com/patternfly/patternfly-react/issues/4602
https://github.com/patternfly/patternfly-react/pull/4603

PR #703, opened 2020-07-22
Update Battery text in PDF

LI0131:
This will update the labels for the battery icons on the advisor side 😁

It will need to go in after https://github.com/RedHatInsights/frontend-components/pull/643 and I update the deps


AllenBW:
@LI0131 yah mind posting ah pdf with this change (in the pr? )

LI0131:
@AllenBW of course 😀 

Screen Shot 2020-07-22 at 8 57 34 AM

PR #704, opened 2020-07-22
Return all data from inventory tags endpoint

LI0131:
This PR doesn't change anything visually just some code changes that I think will help once the /tags endpoint gets fixed by inventory 👍


LI0131:
@AllenBW I updated that grouping method for the tags so 🤞 they should show up now

PR #705, opened 2020-07-22
Add filter chips to Tags Toolbar

LI0131:
This PR adds the config for filter chips on the Tags Modal. This will be ready to go once we get https://github.com/RedHatInsights/frontend-components/pull/645 and the deps updated appropriately. Here's what they look like.

Screen Shot 2020-07-22 at 2 56 49 PM

Here's what happens if you select a lot of different tags.

Screen Shot 2020-07-22 at 2 59 23 PM

### DIFF
--- a/src/PresentationalComponents/TagsToolbar/ManageTags.js
+++ b/src/PresentationalComponents/TagsToolbar/ManageTags.js
@@ -89,7 +89,7 @@ const ManageTags = ({ toggleModal, fetchTags, selectedTags, setSelectedTags, tot
             }))
         }}
         loaded={ loaded }
-        width='auto'
+        width='50%'
         isOpen={ isOpen }
         toggleModal={() => {
             setPerPage(10);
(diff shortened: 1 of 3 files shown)


src/PresentationalComponents/TagsToolbar/ManageTags.js
```diff
@@ -37,11 +56,25 @@ const ManageTags = ({ toggleModal, fetchTags, selectedTags, setSelectedTags, tot
...
+            setSelected([]);
+        } else {
+            const ids = items.flatMap(item => item.chips.map(chip => chip.id));
+            setSelected(selected.filter(tag => ids.filter(id => id !== tag.id).length));
+        }
+    };
+
     return <TagModal
-        systemName={intl.formatMessage(messages.insightsHeader).toLowerCase()}
+        title={intl.formatMessage(messages.tagsCount, { count: totalTags })}
```
AllenBW, line 77: nice followup 🙇 🤗

PR #715, opened 2020-08-04
Don't truncate system table column headings

mhuth:
@AllenBW, my attempt at https://projects.engineering.redhat.com/browse/RHCLOUD-8282.  Not sure if this is the right way to go about fixing this tho and the last_seen column treatment is a bit experimental.  I've hidden it for a bit when the screen width really squeezes the system name column.  Feel free to close this if you've got something else (ie better :smile:) in mind.

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -1,4 +1,6 @@
 /* eslint camelcase: 0 */
+import './SystemsTable.scss';
+
 import * as AppActions from '../../AppActions';
 import * as ReactRedux from 'react-redux';
 import * as pfReactTable from '@patternfly/react-table';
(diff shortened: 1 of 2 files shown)


mhuth:
Full column names appear on all screen sizes, but the Last Seen column is visible and hidden depending on screensize.  Good or :vomiting_face: idea?

2xl_screensize

lg_screensize

md_screensize

terezanovotna:
Hiding the last column is not the greatest idea. 
Is there any other way we can see all the columns but also read the table titles without abbreviating them?

LI0131:
@mhuth you might check out `pfReactTable.fitContent`. I used that for a similar issue on the recs table 🤷‍♂️

terezanovotna:
Screen Shot 2020-08-04 at 15 22 32
I took a screenshot of current abbreviation and there should be enough space to fit it all.

AllenBW:
`.fitContent` might not work here as the numbers are real small (tiny content), n the column header size is what we're lookin' to expand... 🤔

mhuth:
Ok @AllenBW, @LI0131, @terezanovotna, I removed the column hiding idea.  It was pretty :vomiting_face: anyway.  But I've added some CSS that also might be a bit controversial :scream:.  I've decreased the spacer between the heading text and the sorting icon (it was 1rem and now its 0.25rem), and I've set the right padding for the th and td elements to 0px.  These together save a bit more space and still look ok (I think so anyway).  I've kept `pfReactTable.wrappable` as it seems to be about the same as `fitContent`.  Anyway, LMK what you think :thinking: 

space_saving_measures

AllenBW:
That last seen column has some odd spacing... gonna look at it 👁️ n then probably merge this 🤠

AllenBW:
ok commit i pushed up does this
Screen Shot 2020-08-05 at 8 11 30 AM
Screen Shot 2020-08-05 at 8 11 18 AM

PR #725, opened 2020-08-11
Add RemediationButton as dedicatedAction

LI0131:
This PR gives a fix for https://projects.engineering.redhat.com/browse/RHCLOUD-6054. This might not be the desired end fix, but it will make it so the wizard always mounts. It looks like vulnerability has a ticket out for this too.

Before:

Screen Shot 2020-08-11 at 9 18 12 AM

After:

Screen Shot 2020-08-11 at 9 17 41 AM


terezanovotna:
@LI0131 looks good! 

nitpick - the ansible logo should be white to match the white text.

PR #728, opened 2020-08-11
Update RuleDetails for compact mode

LI0131:
Fixes some of the styling for compact mode

Before:

Screen Shot 2020-08-11 at 12 55 08 PM

Screen Shot 2020-08-11 at 12 55 29 PM

After:

Screen Shot 2020-08-11 at 12 54 03 PM
Screen Shot 2020-08-11 at 12 54 29 PM

### DIFF
--- a/src/PresentationalComponents/RuleDetails/_RuleDetails.scss
+++ b/src/PresentationalComponents/RuleDetails/_RuleDetails.scss
@@ -3,6 +3,14 @@ hr {
   margin: var(--pf-global--spacer--md) 0 var(--pf-global--spacer--md) 0;
 }
 
+.ins-c-rule-details__split {
+  display: inline-flex;
+  justify-content: space-around;
+  .pf-l-split__item {
+    position: relative;
+  }
+}
+
 .ins-c-rule-details__stack {
   width: 600px;
 
@@ -61,5 +69,17 @@ hr {
   p {
     margin-top: 0;
     margin-bottom: var(--pf-global--spacer--md);
+  }
+}
+
+@media only screen and (max-width: 768px) {
+  .ins-c-center-text {
+    text-align: left !important;
+  }
+  .ins-c-rule-details__split {
+    display: block;
+    .pf-l-split__item {
+      margin: var(--pf-global--spacer--md);
+    }
   }
 }
--- a/src/SmartComponents/Recs/Details.scss
+++ b/src/SmartComponents/Recs/Details.scss
@@ -1,3 +1,5 @@
+@import '~@redhat-cloud-services/frontend-components-utilities/files/Utilities/_all';
+
 .pageHeaderOverride {
   padding-bottom: 0px;
 }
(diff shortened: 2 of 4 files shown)


src/PresentationalComponents/RuleDetails/_RuleDetails.scss
```diff
@@ -63,3 +71,15 @@ hr {
     margin-bottom: var(--pf-global--spacer--md);
   }
 }
+
+@media only screen and (max-width: 768px) {
```
AllenBW, line 75: does this exist as a pf var? feels like we shouldn't be hardcoding this value 🤔

AllenBW: yep
`--pf-global--breakpoint--md | global_breakpoint_md | 768px`

src/SmartComponents/Recs/Details.scss
```diff
@@ -17,3 +19,9 @@
 .categoryLabel {
   margin-left: var(--pf-global--spacer--md)
 }
+
+@media only screen and (max-width: 768px) {
```
AllenBW, line 23: following comment above, set this this to `var(--pf-global--breakpoint--md)`

src/PresentationalComponents/RuleDetails/_RuleDetails.scss
LI0131: So did a little digging on this... and it looks like it isn't possible use `var()` in a media query. https://stackoverflow.com/questions/40722882/css-native-variables-not-working-in-media-queries (unless we want to use a preprocessor?)

AllenBW: oooooo  what about `@media only screen and (min-width: $ins-break--md) `

AllenBW: actually you dont neet to do that `$ins-break--md: var(--pf-global--breakpoint--md);` is already defined in
@import '~@redhat-cloud-services/frontend-components-utilities/files/Utilities/_all';

LI0131: Oooo I had the same thought 😂 

I think that approach falls victim to the same issue though. The styles don't seem to apply 🤷‍♂️🤔

AllenBW: local definition as well?  🤔 I see the value hardcoded (not the pf var) in a few places.. would hate for us to also do it 😢 will take a look tomorrow 👍

AllenBW: great minds 🧠 😆

AllenBW: well... s on the radar https://github.com/w3c/csswg-drafts/issues/2627 but yeah not currently possible in the way   would hope

PR #743, opened 2020-08-31
Removes 'undefined' as reports_shown url param (when applicable)

AllenBW:
related to https://github.com/RedHatInsights/insights-advisor-api/pull/631

internally we need to track the undefined state, but visually, n as far as the url is concerned no need to broadcast this invalid url param

furthermore, if redirect to with the param attribute missing or undefined, will behave as expected (current state) will show all status of recs

#### looks like
Screen Shot 2020-08-31 at 11 34 53 AM
Screen Shot 2020-08-31 at 11 34 43 AM


dkuc:
:+1:

PR #748, opened 2020-09-08
Fixes global filter tagging, removes hardcoded chrome source

AllenBW:
follow on to#747 that fixed very little

#### success!
Screen Shot 2020-09-08 at 9 57 23 AM


LI0131 (approved):
:godmode: 💯

PR #749, opened 2020-09-09
Set flex value for Recs header

LI0131:
This PR addresses https://projects.engineering.redhat.com/browse/RHCLOUD-8707

**Before:**
Screen Shot 2020-09-09 at 2 52 36 PM

**After:**

1. Safari
Screen Shot 2020-09-09 at 2 43 38 PM

2. Chrome
Screen Shot 2020-09-09 at 2 44 18 PM

3. Firefox
Screen Shot 2020-09-09 at 2 43 22 PM

PR #750, opened 2020-09-10
Keep Rec Header size consistent

LI0131:
This PR reverts #749 in order to keep the Rec Header size consistent when users enter text into the search field. 

Addresses 🎫  -> https://projects.engineering.redhat.com/browse/RHCLOUD-9063

Before:
Screen Shot 2020-09-09 at 8 32 32 PM

After:
Screen Shot 2020-09-09 at 8 33 50 PM


AllenBW (approved):
👍

PR #760, opened 2020-09-30
Only display Red Hat disabled rules alert if there are said rules

mhuth:
Does it make sense to announce about Red Hat Disabled rules if there aren't any rules just yet?  At the moment I don't believe we have any Red Hat disabled rules and not sure when that'll start happening.  So (to me at least) it seems better to make the alert conditional until we have such rules and the customer can be made aware of them then.

### DIFF
--- a/src/SmartComponents/Recs/List.js
+++ b/src/SmartComponents/Recs/List.js
@@ -1,8 +1,9 @@
 import './List.scss';
 
 import { PageHeader, PageHeaderTitle } from '@redhat-cloud-services/frontend-components/components/PageHeader';
-import React, { Suspense, lazy, useState } from 'react';
+import React, { Suspense, lazy, useState, useEffect } from 'react';
 
+import API from '../../Utilities/Api';
 import { Alert } from '@patternfly/react-core/dist/esm/components/Alert/Alert';
 import { AlertActionCloseButton } from '@patternfly/react-core/dist/esm/components/Alert/AlertActionCloseButton';
 import DownloadExecReport from '../../PresentationalComponents/ExecutiveReport/Download';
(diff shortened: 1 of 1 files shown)


mhuth:
If there are no Red Hat disabled rules, then don't display the alert:
no_rules_no_alert

Only when there are Red Hat disabled rules is the alert shown:
alert_when_rules

Of course if the customer clicks on the X then then alert won't be shown either.

terezanovotna:
@mhuth this makes a lot of sense. I like this proposal👍

terezanovotna (approved):

src/SmartComponents/Recs/List.js
```diff
@@ -20,13 +21,26 @@ let AdvisorRedHatDisabledRuleAlert = sessionStorage.getItem('AdvisorRedHatDisabl
...
+    const [redHatDisabledRuleCount, setRedHatDisabledRuleCount] = useState(0);
+
+    useEffect(() => {
+        (async () => {
+            try {
+                const response = (await API.get(`${RULES_FETCH_URL}?rule_status=rhdisabled`)).data;
+                setRedHatDisabledRuleCount(response.meta.count);
+            } catch (error) {
+                setRedHatDisabledRuleCount(0);
+                throw `${error}`;
```
AllenBW, line 33: m thinking lets not throw an error over this fail let it be silent n just not show the message? 🤔

src/SmartComponents/Recs/List.js
```diff
@@ -20,13 +21,26 @@ let AdvisorRedHatDisabledRuleAlert = sessionStorage.getItem('AdvisorRedHatDisabl
...
+        })();
+    }, []);
 
     return <React.Fragment>
         {!isGlobalFilter() && <Suspense fallback={<Loading />}> <TagsToolbar /> </Suspense>}
         <PageHeader className='ins-c-recommendations-header'>
             <PageHeaderTitle title={`${intl.formatMessage(messages.insightsHeader)} ${intl.formatMessage(messages.recommendations).toLowerCase()}`} />
             <DownloadExecReport />
-            {redhatDisabledRuleAlert &&
+            {redhatDisabledRuleAlert && redHatDisabledRuleCount > 0 &&
```
AllenBW, line 43: `redHatDisabledRuleCount.length` would beeeeee good herereereer

AllenBW (commented):
LOVE THIS

src/SmartComponents/Recs/List.js
AllenBW: on second thought, imma go back to sleep

src/SmartComponents/Recs/List.js
mhuth: Done and done!

AllenBW (approved):
🥎 🏆 🔔

PR #763, opened 2020-10-02
Filter out non-filterTypes from chips 

LI0131:
This PR fixes https://issues.redhat.com/browse/ADVISOR-1387

Right now we get an extra `Name` filter chip when `reports_shown` is found in the URL.
https://cloud.redhat.com/insights/advisor/recommendations?impacting=true&sort=-total_risk&limit=10&offset=0&total_risk=3&reports_shown=true

**Before:**
Screen Shot 2020-10-02 at 10 21 13 AM

**After:**
Screen Shot 2020-10-02 at 10 21 27 AM


AllenBW (approved):
best thing since... sliced... 🍞 


💯 :shipit: 🧟

PR #766, opened 2020-10-14
No Recommendations text message reflects the selected rule status filter

mhuth:
Looking at the text for 'Disabled' and 'Red Hat disabled' when there are no recommendations - "None of your connected systems are affected by any known rules." - this doesn't seem to accurately convey why there are no recommendations for 'Disabled' and 'Red Hat disabled'.  This PR changes the text to more accurately reflect why there might be no recommendations for these 2 status options.

When there are 'No Recommendations' for Status of Disabled:
no_disabled_recommendations

When there are No Recommendations for Status of Red Hat disabled:
no_redhat_disabled_recommendations

What do y'all think?  Feel free to shoot me down in flames if you think the old messaging was fine.


terezanovotna:
thanks for thinking about it @mhuth!

Both sentences make sense, but how about we make them even more consistent? 
For disabled:
Currently there are no disabled recommendations.

mhuth:
> thanks for thinking about it @mhuth!
> 
> Both sentences make sense, but how about we make them even more consistent?
> For disabled:
> Currently there are no disabled recommendations.

Yep, can do @terezanovotna .  My motivation for including the 'you' in the disabled recommendations message was to indicate that this is something that the customer controls.  But they don't control the Red Hat disabled recommendations.

AllenBW (dismissed):
Can we real quick get an update to compiled translations `npm run translations` 🤗

mhuth:
Done.  Sorry, I always forget about the translations :/

AllenBW:
S easy to forget, 🤗 🥇

AllenBW (approved):
:godmode:

PR #768, opened 2020-10-19
Fix styling of autoacks alert on small screen sizes

mhuth:
Noticed the styling of the Autoack alert was mangled on smaller screen sizes.  This PR tries to make it purty(er).  Please LMK if there is a better / more-patternfly-ish way of doing this.

Before style fix:
before

After style fix:
after

At the breakpoint of style fix:
at_breakpoint

### DIFF
--- a/src/PresentationalComponents/ExecutiveReport/_Download.scss
+++ b/src/PresentationalComponents/ExecutiveReport/_Download.scss
@@ -5,4 +5,8 @@
 
 .downloadButtonOverride {
     float: right;
+    @media only screen and (max-width: 585px) {
+        float: none;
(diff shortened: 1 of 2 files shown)


terezanovotna (approved):

AllenBW (approved):
what. who is using crc at ah smaller than 600px screen! yer their savior, looks good to me!

mhuth:
> what. who is using crc at ah smaller than 600px screen! yer their savior, looks good to me!

Heh, yeah I was just trying to fix this (the text not wrapping in the alert) and went a bit nuts with the small screen stuff :smile: 
image (1)

PR #769, opened 2020-10-19
Add Redirect component for Classic ids

LI0131:
Resolves https://issues.redhat.com/browse/ADVISOR-1403 by creating a component which handles the api request for the inventory id corresponding to the given classic id and redirects to `InventoryDetails`.

When redirect fails:

Screen Shot 2020-10-20 at 8 13 47 AM

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -979,5 +979,10 @@ export default defineMessages({
         id: 'redhatDisabledRuleAlertTitle',
         description: 'Red Hat disabled rule alert title',
         defaultMessage: 'Red Hat has proactively disabled certain low risk recommendations'
+    },
+    inventoryIdNotFound: {
+        id: 'inventoryIdNotFound',
+        description: 'Thrown as error when classic id does not correspond to an inventory id',
+        defaultMessage: 'No system found in inventory for the given Advisor ID'
     }
 });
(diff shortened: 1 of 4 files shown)


AllenBW (changes requested):
Can yah npm run translations 🤞

AllenBW (dismissed):
Really like this, nicely done @LI0131 ! very self contained, all good pending ux 👌

src/Messages.js
```diff
@@ -979,5 +979,10 @@ export default defineMessages({
         id: 'redhatDisabledRuleAlertTitle',
         description: 'Red Hat disabled rule alert title',
         defaultMessage: 'Red Hat has proactively disabled certain low risk recommendations'
+    },
+    inventoryIdNotFound: {
+        id: 'inventoryIdNotFound',
+        description: 'Thrown as error when classic id does not correspond to an inventory id',
+        defaultMessage: 'No system found in inventory for the given Insights ID'
```
AllenBW, line 986: should this say Advisor ID instead ?  🤔

LI0131: I was going off of `/api/inventory/v1/hosts?insights_id=...` but we could change it to that 🤷‍♂️

AllenBW: Yeah lets change that, ux might say differently, but Advisor is the product (in the Insights portfolio?), lil more meaningful methinks 🤔

terezanovotna:
Looking good!

AllenBW (approved):
🙇‍♂️🥰✅

PR #770, opened 2020-10-20
Properly persist clearing the Red Hat disabled rule alert

mhuth:
Actually works this time.   Whodathunk 'false' === true !?!  Programming101 fail for Mark :facepalm: 

But anyway, should I use localStorage instead of sessionStorage though?  Then the alert will never be back :terminator:

### DIFF
--- a/src/SmartComponents/Recs/List.js
+++ b/src/SmartComponents/Recs/List.js
@@ -16,10 +16,9 @@ import { useIntl } from 'react-intl';
 const RulesTable = lazy(() => import(/* webpackChunkName: "RulesTable" */ '../../PresentationalComponents/RulesTable/RulesTable'));
 const TagsToolbar = lazy(() => import(/* webpackChunkName: "TagsToolbar" */ '../../PresentationalComponents/TagsToolbar/TagsToolbar'));
 
-let AdvisorRedHatDisabledRuleAlert = sessionStorage.getItem('AdvisorRedHatDisabledRuleAlert') || true;
-
 const List = () => {
     const intl = useIntl();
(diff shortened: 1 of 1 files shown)


LI0131 (commented):
Looking good @mhuth 😁 You might consider using `JSON.parse` to deserialize the value from the sessionStorage to boolean, but strings will definitely get it done 👍

src/SmartComponents/Recs/List.js
```diff
@@ -39,14 +37,14 @@ const List = () => {
         <PageHeader className='ins-c-recommendations-header'>
             <PageHeaderTitle title={`${intl.formatMessage(messages.insightsHeader)} ${intl.formatMessage(messages.recommendations).toLowerCase()}`} />
             <DownloadExecReport />
-            {redhatDisabledRuleAlert && redHatDisabledRuleCount > 0 &&
+            {redhatDisabledRuleAlert === 'true' && redHatDisabledRuleCount > 0 &&
```
AllenBW, line 40:
```suggestion
            {redhatDisabledRuleAlert && redHatDisabledRuleCount > 0 &&
```

agree with @LI0131 makin these bools means we can simplify this line

AllenBW (approved):
B.E.A.U.T.I.F.U.L.

src/SmartComponents/Recs/List.js
mhuth: Done!  Yep, bools look much better than strings.

mhuth:
> Looking good @mhuth  You might consider using `JSON.parse` to deserialize the value from the sessionStorage to boolean, but strings will definitely get it done

Ah of course!  Thanks for the tip @LI0131

LI0131 (approved):
Awesome @mhuth 😁

PR #771, opened 2020-10-20
Update ClassicRedirect for urls with beta

LI0131:
I noticed a bug with the classic redirect when in the beta environments. I hardcoded the indices we are using to determine the `id` and `classicId` values, and they become set to the wrong values, because the indices I used do not account for `/beta`

For URL: https://cloud.redhat.com/beta/insights/advisor/recommendations/classic/decreased_vm_network_performance%7CVHOST_NET_KERNEL_MODULE_NOT_LOADED/1ffb3a80-8290-4b2f-ba9d-9e6422604fd6
Screen Shot 2020-10-20 at 9 25 02 AM

### DIFF
--- a/src/SmartComponents/Recs/ClassicRedirect.js
+++ b/src/SmartComponents/Recs/ClassicRedirect.js
@@ -16,8 +16,9 @@ const ClassicRedirect = () => {
     const intl = useIntl();
     const dispatch = useDispatch();
     const pathname = window.location.pathname.split('/');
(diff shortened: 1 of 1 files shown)


AllenBW (approved):
In this case we dont' need to worry about beta, cuz.... satellite wont ever redirect to beta (ever) ((and if they do one day, i'll eat a beansprout))

THAT BEING SAID never one to turn down a lil good code 👍

PR #772, opened 2020-10-21
Use paramsSerializer for tag param parsing, update deps 😀😃😄

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1412

similar to https://github.com/RedHatInsights/insights-dashboard/pull/202


src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -73,7 +73,7 @@ const RulesTable = ({ rules, filters, rulesFetchStatus, setFilters, fetchRules,
 
     const fetchRulesFn = useCallback(() => {
         urlBuilder(filters, selectedTags, workloads);
-        let options = selectedTags.length && ({ tags: selectedTags.map(tag => encodeURIComponent(tag)).join('&tags=') });
```
LI0131, line 76: @AllenBW not sure if this an issue or not 🤔 but it looks like the tags are getting joined with `,` instead of with `&tags=` in this implementation

AllenBW: yeah they are, not an issue, api supports it both ways, cyndi enables support for &tags=

LI0131: well that's awesome 😁 💯

LI0131 (approved):
Awesome stuff @AllenBW 🥇

PR #778, opened 2020-10-30
Update ClassicRedirect to consume systems/classic

LI0131:
This PR adds additional functionality to the `ClassicRedirect` component. Using a switch statement we can now handle both `recommendations/classic` and `systems/classic`.

Resolves: https://issues.redhat.com/browse/ADVISOR-1455


AllenBW (approved):
yes.

PR #784, opened 2020-11-21
Add federated modules generation and custom plugins

karelhala:
**Very much work in progress, do not merge!**

### Chrome 2.0

Since chrome wants to use SPA and migrate to start using federated modules to load chunks of UI we want to adapt to it by following the guide for it [migration guide](https://github.com/RedHatInsights/insights-chrome/blob/master/docs/migration-guide.md).

### TODO:
* [x] upgrade to new version of shared config
* [x] start using federated modules
* [x] connect to new inventory
* [x] plug into chrome's plugin system
* [x] make sure nothing breaks.


Hyperkid123:
inventory routing is fixed by: https://github.com/RedHatInsights/insights-chrome/pull/1082

karelhala:
@AllenBW everything has been updated, it should work both in legacy and new chrome without any error. Nothing should change from dev experience, you can run the usual tasks and it should work out of the box. You'll just have to reinstall all deps to properly build the app with new webpack and shared config.

We'll keep the config in beta for some time, until we know nothing breaks. Once we are sure, we'll release new major version since this is completely changing the way how things are built. If you are interested in the config changes here is the PR https://github.com/RedHatInsights/frontend-components/pull/813

If you have any problems with running it or hot reload or anything let us know.

AllenBW:
Ok will jump in n play around!! if s all good, we'll push to beta and 🤞

AllenBW:
@karelhala so cleaned `./node_modules` did an npm i proxy running as normal, npm start... all i get is `Cannot GET /insights/advisor/recommendations/` refresh shows action on the proxy term but nothing on the serve term 🤔

karelhala:
@AllenBW the problem is in the mismatched version of predefined webpack config and webpack itself. In WP5 they forced one config value, but with the newest version it's no longer needed. We've updated our shared config to match the newest version of webpack.

There's also a small problem with inventory component and refreshes, however that is being addressed in https://github.com/RedHatInsights/frontend-components/pull/881 once that is merged I think we are golden!

Hyperkid123:
> ./node_modules did an npm i proxy running as normal, npm start... all i get is Cannot GET /insights/advisor/recommendations/ refresh shows action on the proxy term but nothing on the serve term thinking

Updating webpack dependencies to the latest fixes the issue. We still have an older dependency version in the config package.

AllenBW:
Ya'll are too awesome 🤗

AllenBW:
WOOOOH there she goesssss only two issues:

- local styles aren’t being applied, (ansible check color is black, rather than green as specified in the _RulesTable.scss imported by the component), updating to css-loader 5.0.1 didn't seem to fix... wondering if yah have any thoughts @Hyperkid123 ? 



- system detail view is borked ```Warning: Failed prop type: Invalid prop `component` of type `object` supplied to `InsightsRoute`, expected `function`.```  not sure if thats the remediation component thing showing up again or not

<3

Hyperkid123:
@AllenBW 
> local styles aren’t being applied, (ansible check color is black, rather than green as specified in the _RulesTable.scss imported

That is because we are scoping styles to avoid potential global app styling conflicts (you know somebody adds .pf-c-tile styles). We need to add a top-level class name which is the same as the `insights.app` value in `package.json`. Chrome 2 will do it on its own but we are not there yet. cc @karelhala

Hyperkid123:
@AllenBW so because the advisor is using the insights route and mutates the root element (Which it no longer should do in chrome 2), adding the class at build time is not enough. I've added the required class name to the insights route.

For applications that are not mutating the root element: https://github.com/RedHatInsights/frontend-components/pull/813/commits/73e49bf2d330ce33d97e8ab538ff442e641af341

But the styling issues should no longer be an issue.

AllenBW:
Style be STYLISH!! just that last piece, the system details page being a lil broken is holding this one from promise land of beta 🥇

AllenBW:
@Hyperkid123

Hyperkid123:
Well. I have absolutely no idea what is causing the issue then... At least it's working for you :smile:.  Fun debugging session ahead!

AllenBW:
If yah wanta jump on ah 👖 n 👀 lemme know!

Hyperkid123:
It's alright we will get it working.

AllenBW:
Any chance theres n update on the broken system details page?

karelhala:
@AllenBW sorry for the late reply, There were actually 2 problesm:
* old version of FEC which didn't had fallback for `onLoad`
* old version of remediations package which didn't included the remediations button properly

AllenBW:
@karelhala no worries! THIS LOOKS FANTASTIC all things full steam aheaddddd OFF TO BETA 🚢 :shipit: 🚂 🌬️

AllenBW (approved):
yes. 🆗

PR #785, opened 2020-11-29
Store the alert close action in local storage (.. and update deps)

mhuth:
Fixes https://issues.redhat.com/browse/ADVISOR-1496


AllenBW (approved):
Not sure how the lockfile got fussied up, rebuild n pushed, lets get this onto beta stat!

PR #789, opened 2021-01-06
Remove Red Hat from insights

karelhala:
### Remove Red Hat from insights

Since the branding has changed slightly we want to remove the words `Red Hat` from browser title as well.


AllenBW (approved):
😎🆒

PR #790, opened 2021-01-07
Make sure impacting is true when rule_status is enabled

mhuth:
Fixes https://issues.redhat.com/browse/ADVISOR-1532 ... I think.

Was able to reproduce issue.  Go to recommendations page and export data is enabled by default because rule_status = enabled and impacting is true.  Change rule status to disabled (or anything other than enabled) and export data is disabled ... all good so far.  But then change rule_status back to enabled and the export data button is still disabled.

Before patch:
before

After patch:
after

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -101,7 +101,7 @@ const RulesTable = ({ rules, filters, rulesFetchStatus, setFilters, fetchRules, 
     };
 
     const toggleRulesDisabled = (rule_status) => {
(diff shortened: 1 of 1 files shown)


AllenBW:
Tagging @terezanovotna fore review, looks fine ish code wise, but worried about shadow manipulating other filter values, thinking the whole function of the kebab should be rearranged

terezanovotna:
looks good. Export button should not be disabled.

AllenBW:
This action was change here... https://github.com/RedHatInsights/insights-advisor-frontend/pull/498

"re-enabling the export button " isn't the quick fix we're looking for

AllenBW:
closin this for now, pending unanimous decision

PR #796, opened 2021-02-15
Federated modules debugging

Hyperkid123:
@AllenBW I've made a little clean up after we used advisor for the chrome 2 testing. Also removed some legacy code that is not required anymore.


AllenBW (approved):
WONDERFUL! thanks for this, works like ah charm

PR #798, opened 2021-02-16
Adds support for pdf export group names

AllenBW:
fixes https://issues.redhat.com/browse/RHCLOUD-12429

#### Now includes "Red Hat Insights" title
Screen Shot 2021-02-16 at 11 36 55 AM
Screen Shot 2021-02-16 at 11 33 59 AM

### DIFF
--- a/locales/translations.json
+++ b/locales/translations.json
@@ -105,6 +105,7 @@
   "recommendation": "Recommendation",
   "recommendations": "Recommendations",
   "recommended": "Recommended",
+  "redHatInsights": "Red Hat Insights",
   "redhatDisabled": "Red Hat disabled",
   "redhatDisabledRuleAlert": "We want you to focus on the more important risks to your systems. Red Hat disabled recommendations can be found under the Status filter.",
   "redhatDisabledRuleAlertTitle": "Red Hat has proactively disabled certain low risk recommendations",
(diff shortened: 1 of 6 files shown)


mhuth (approved):
Seems simple enough I don't even have to fetch and run it :+1: (is that being irresponsible?)

terezanovotna:
Instead of adding a title "Red Hat Insights", can we change the logo? Check out the example what we are doing for emails https://marvelapp.com/prototype/26g06806/screen/76353794 If we had Red Hat Insights within the logo, we don't need to mention it in the title again

AllenBW:
@terezanovotna cant so much change the logo at this time, this work was more so to align with vul ui https://github.com/RedHatInsights/frontend-components/pull/968

terezanovotna:
Well. I am ok if we merge it - to be consistent. But the next step, I would love to see the new logo applied to all applications that do export. cc @katierik (seems like our design has evolved and my suggestion is to include _Red Hat Insights logo_ in the report instead of only _Red Hat logo_)

katierik:
Good call @terezanovotna - Can you grab a HQ copy of the logo and open bugs against the reports? I think we have reports for Advisor, Vulnerability, & Compliance. (cc me plz!)

e.g. in compliance we don't have the line "Red Hat Insights" but it does say "Insights Compliance" -- in this case I think we would drop "Insights" and just rely on the logo.
https://marvelapp.com/prototype/6bdbgig/screen/76389793

PR #800, opened 2021-03-01
Reconfigures impacted systems kebab to be a filter, removes asyncComp 🍔💀

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1591

mocks https://marvelapp.com/prototype/26g06806/screen/76713877

small caveat, when selecting none, setting impacting=false, api does not filter by only those recs without system hits, it returns all systems @mhuth @PaulWay we might want to change this? maybe maybe? idk 🤷 

#### looks like thissssss
Screen Shot 2021-03-01 at 12 58 25 PM
Screen Shot 2021-03-01 at 12 55 43 PM
Screen Shot 2021-03-01 at 12 55 34 PM
Screen Shot 2021-03-01 at 12 55 25 PM


mhuth:
Looks good @AllenBW but I must admit the impacting parameter is confusing to me as well and is complicated in the API and is doing some voodoo magic that @PaulWay understands much better than I.  We might even need to create a new parameter to make this 'systems impacted' thing work as expected.

AllenBW:
@mhuth we'll talk about it after sprint planning! or during 🤷

terezanovotna:
looks good!

PR #802, opened 2021-03-08
Fix rules concurrency

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-12765

There is an issue with calling the `fetchRules` actions. The problem is the following. The app creating multiple rule requests upon changing the global filter value. The global filter change also triggers local filter changes. That causes two calls to be made. The local filter call is made first and contains the exact same parameters as the previous request. So it returns the exact same payload. Then the correct API is made. If the first promise takes longer (which is very common), it overrides the correct data and we have stale information stored in the redux.

This bug is present even on prod stable environment.

This PR creates a promise queue for the fetch rules redux action and always returns the last promise response. This is a hot fix, not something that should be used for a long time. The filtering logic has to be either fixed or re-done. I don't have enough knowledge about the advisor codebase. I will leave it to you @AllenBW 

Screenshot from 2021-03-08 13-30-39


AllenBW (approved):

AllenBW:
merging for now, will do more work when back 🤗 ♾️

PR #803, opened 2021-03-15
Removes local-proxy requirement, update deps 

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1624


config/dev.webpack.config.js
```diff
@@ -4,7 +4,20 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPl
 const config = require('@redhat-cloud-services/frontend-components-config');
 const { config: webpackConfig, plugins } = config({
     rootFolder: resolve(__dirname, '../'),
-    debug: true
+    debug: true,
+    useProxy: process.env.USEPROXY ? true : false,
+    deployment: process.env.BETA ? 'beta/apps' : 'apps',
```
Hyperkid123, line 9: This may break your build, you want to keep the deployment `undefined` if not using the beta env variable. Like here: 
https://github.com/RedHatInsights/frontend-starter-app/blob/master/config/dev.webpack.config.js#L8.

Also, you are missing the start:beta command: https://github.com/RedHatInsights/frontend-starter-app/blob/master/package.json#L19

AllenBW: As this is only the dev webpack config, build should be fine right?

config/dev.webpack.config.js
```diff
@@ -4,7 +4,19 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPl
...
+    customProxy: process.env.USEPROXY ? [
+        {
+            context: (path) => path.includes('/api/'),
+            target: 'https://cloud.redhat.com',
+            secure: true,
+            changeOrigin: true,
+            autoRewrite: true,
+            ws: true
+        }
+    ] : []
```
karelhala, line 19: This only referes to prod api, I think you could have something like
```JS
customProxy: process.env.API_ENDOINT ? [
        {
            context: (path) => path.includes('/api/'),
            target: process.env.API_ENDOINT,
            secure: true,
            changeOrigin: true,
            autoRewrite: true,
            ws: true
        }
    ] : []
```

And when running the app `API_ENDOINT=https://cloud.redhat.com npm start` or for CI api `API_ENDOINT=https://ci.cloud.redhat.com npm start`

package.json
```diff
@@ -18,7 +18,7 @@
     "nightly": "npm run travis:verify",
     "prod": "NODE_ENV=production webpack serve --config config/dev.webpack.config.js",
     "server:ctr": "node src/server/generateServerKey.js",
-    "start": "NODE_ENV=development webpack serve --config config/dev.webpack.config.js",
+    "start": "NODE_ENV=development USEPROXY=true webpack serve --config config/dev.webpack.config.js",
```
karelhala, line 21: This will force using the proxy, do you really want it? I guess ading new z`start:proxy` might be a better approach.

AllenBW: mmm well i mention in the readme removing that allows all original functionality...  but yeah adding a new script is probs even more kind

config/dev.webpack.config.js
AllenBW: good point! tbh, i dont want to type that out every time, but encapsulating this mod with yer second comment would allow for maximum flexibility, thanks for lookin @karelhala 🤗

config/dev.webpack.config.js
```diff
@@ -4,7 +4,19 @@ const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPl
 const config = require('@redhat-cloud-services/frontend-components-config');
 const { config: webpackConfig, plugins } = config({
     rootFolder: resolve(__dirname, '../'),
-    debug: true
+    debug: true,
+    useProxy: process.env.USEPROXY ? true : false,
+    localChrome: process.env.INSIGHTS_CHROME,
+    customProxy: process.env.USEPROXY ? [
```
karelhala, line 10: Shouldn't this be `API_ENDOINT`?
```suggestion
    customProxy: process.env.API_ENDOINT ? [
```

package.json
```diff
@@ -19,6 +19,7 @@
     "prod": "NODE_ENV=production webpack serve --config config/dev.webpack.config.js",
     "server:ctr": "node src/server/generateServerKey.js",
     "start": "NODE_ENV=development webpack serve --config config/dev.webpack.config.js",
+    "start:proxy": "NODE_ENV=development USEPROXY=true webpack serve --config config/dev.webpack.config.js",
```
karelhala, line 22: It would probably be better to use the prod proxy here
```suggestion
    "start:proxy": "NODE_ENV=development API_ENDOINT=https://cloud.redhat.com webpack serve --config config/dev.webpack.config.js",
```

config/dev.webpack.config.js
AllenBW: nope, cuz we only want to use the custom proxy if USEPROXY is set, default target API without is crc ya'd do `API_ENDOINT=https://ci.cloud.redhat.com npm run start:proxy` to specify the proxy

package.json
AllenBW: okie dokie, relied on api_endpoint as  useproxy flag, start:proxy now by default runs prod, `API_ENDOINT=https://ci.cloud.redhat.com npm run start` would be a custom proxy

karelhala (approved):

PR #810, opened 2021-04-13
Overhaul system table, global filter renders sap_system param irrelevant

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1634

@hyperkid123 OK so this is fixed in chrome 2.0, ie beta, BUT right now... if you go directly to advisor/systems... yah don't get the correct rows... but upon changing something like the global filter or navigating away n then back, they show up... was gonna try n fix it for stable in this pr... but then didn't... cuz i couldn't figure it out (yet?)

well looks like we're making this a mammoth n also switching linting


Hyperkid123:
@AllenBW reason why it's **probably** not working in stable env is that there was an ancient bug inventory which was preventing the table update. That has been fixed recently. We still plan to release chrome 2 for the summit so it will fix itself by then.

AllenBW:
@Hyperkid123 yah figured, s why m not sweating it too much, thanks for sayin! 🤗

PR #817, opened 2021-04-28
Ensures no right margin on last filter chip item 🍩☠︎

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1618

also i know that sys-table text is incorrect, we already changed it samezies as the rec table, i might be that the inventory table doesn't accept the `deleteTitle` prop on the `activeFiltersConfig` ? 

Screen Shot 2021-04-28 at 9 12 54 AM
Screen Shot 2021-04-28 at 9 12 40 AM

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,21 +1,35 @@
 // Importing Global Variables
-@import '~@redhat-cloud-services/frontend-components-utilities/styles/variables';
+@import "~@redhat-cloud-services/frontend-components-utilities/styles/variables";
 
 // Temp nav styles
 ul.navigation-secondary {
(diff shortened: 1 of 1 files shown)


src/App.scss
```diff
@@ -1,21 +1,35 @@
...
+      background: #f5f5f5;
+      color: var(--pf-global--link--Color);
+    }
+    &:hover {
+      color: var(--pf-global--link--Color);
+    }
+  }
+}
+
+// Style overrides
```
AllenBW, line 28: everything else is just reformatting, this n below is the work

PR #818, opened 2021-05-04
Move system detail from FEC to this application

karelhala:
### Use federated modules to share system detail

Since we can now use federated modules in every environment let's share the system detail with it. This way we don't have to maintain the version of it in all environments, it's the same for all users on one environment. If a bug is found the fix is deployed just in this application and propagated to inventory via federetad modules.

### Usage
```JSX
import React from 'react';
import { useStore } from 'react-redux';
import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';

const AdvisorTab = () => {
    return <AsyncComponent
        appName="advisor"
        module="./SystemAdvisor"
        store={useStore()}
        customItnl
    />;
};
```


AllenBW (approved):
Okie dokie with this soon as travis is green, excited to rewrite the old version of react away 😜

AllenBW (changes requested):
Nope i take that bad, wanta do the rewrite in this instead of somewhere else, will be throwing commits against this pr shortly

config/dev.webpack.config.js
```diff
@@ -34,9 +34,9 @@ plugins.push(
       root: resolve(__dirname, '../'),
       exposes: {
         './RootApp': resolve(__dirname, '../src/AppEntry'),
-        './SystemDetail': resolve(
+        './SystemAdvisor': resolve(
```
karelhala, line 37: I wouldn't rename the exported module name, we can keep the `./SystemDetail`. After all it's a system's detail, isn't it? You can keep the `src/SmartComponents/SystemAdvisor` file name though.

karelhala:
@AllenBW looking good! Let's just have a consensus on the exported module. We can keep the name of the file, but I wouldn't change the module name. I am aiming to have all apps to expose `./SystemDetail` module. So importing it would be easy
```JS
<AsyncComponent
        appName="advisor" // if I'd want to pull from vulnerability I'd use `vulnerability` same for compliance and patch.
        module="./SystemDetail"
        store={useStore()}
        customItnl
    />;
```

config/dev.webpack.config.js
AllenBW: ah! gottcha on the logic why, onboard with that, will change it back

AllenBW (approved):
s not broken 🤯 s even 🧈 than 🐝 4️⃣

AllenBW:
this work will also fix https://issues.redhat.com/browse/ADVISOR-1719 my thinkin was... why add old code when yah can do a lil work to add new code 🤷

PR #820, opened 2021-05-04
Disable default columns in inventory table

rvsia:
cc @AllenBW 

This ensures that nothing got broken when a new inventory is pushed in chrome.


AllenBW (approved):
<3 thanks for this @rvsia  🌮

PR #824, opened 2021-05-26
Add custom intl prop so the system detail component won't fail

karelhala:
There is an error, after the system detail refactoring the detail was failing on missing Intl. This PR fixes such issue by wrapping the base system detail in wrapper that has the intl provider.


rvsia (approved):

Hyperkid123 (approved):

dkuc (approved):

AllenBW:
pushed to prod-stable

PR #833, opened 2021-06-15
Add default for undefined items

rvsia:
Maybe fixes https://issues.redhat.com/browse/ADVISOR-1723

Items are not required and they can be empty, however there are several `items.length` calls.

There are two approaches how to fix it:

1) add default value `[]`
2) add safe accessors 

Let me know, what you prefer. I went with the first one as it makes more sense to me.


AllenBW (approved):
LG2M, not seeing the issue locally (error loading component) thanks for the pr @rvsia  🤗 🙇

PR #834, opened 2021-06-16
Updates filter wording as per UX request

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1802

#### looks like
Screen Shot 2021-06-16 at 10 47 11 AM


terezanovotna:
Looks great, thanks for tackling it so quickly!

PR #837, opened 2021-06-21
Fixes systems page infinity loading issue

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1803

making a real mess here, might need help

@Hyperkid123 is gonna ❤️‍🔥🦤🔪 me

PR #840, opened 2021-06-23
Fixes systems page infinity loading issue

Hyperkid123:
fixes data loading issue in: https://github.com/RedHatInsights/insights-advisor-frontend/pull/837

@AllenBW feel free to steal the commit and close this PR or the other way around. However, there is something strange with the insights disabled column:

screenshot-prod foo redhat com_1337-2021 06 23-11_36_06


AllenBW:
Ok so the table now displays the correct date... but we still get the stale warning rather than just the last seen tooltip, i can do custom columns n just display the string (maybe grab the [tooltip component here](https://github.com/RedHatInsights/frontend-components/blob/f6d32864b18c85fdf6b12e6025b9f29cbd1d67eb/packages/inventory/src/components/detail/FactsInfo.js)) ? but that seems to be a bit much

AllenBW (approved):
lockfile unrelated to this but moving it in anyways

getEntities appears to break  the  last seen column, removing it fixes this issue, furthermore the work here appears to resolve the infinite loading in chrome

Screen Shot 2021-06-24 at 11 32 45 AM


commenting out the offending code, will revisit in the near future

PR #843, opened 2021-06-28
fun timez making getEntities workin

AllenBW:
Update to sys table to rely on getEntities n remove all the other crutches

everything work EXCEPT when yah input display name n refresh the page it doesn't carry over (everything else does) gotta figure out how to ensure that thing plugs into the inventory table, that aside getting this up there real fast cuz it fixes some noticeable problems


rvsia (approved):
Looks good to me! 🏆

Hyperkid123 (approved):

PR #861, opened 2021-08-16
Make RuleDetails fed module compatible with OCP rules

gkarat:
Check the following sub-components:
- [x] Rule header
- [x] Rule rating
- [x] Impacted systems/clusters (non-detailsPage variant)
- [x] Total risk
- [x] Likelihood line + tooltip message
- [x] Impact line + tooltip message
- [x] Risk of change
- [x] Reboot required message


src/PresentationalComponents/RuleDetails/RuleDetails.js
```diff
@@ -64,15 +67,28 @@ const BaseRuleDetails = ({
       </span>
     );
 
+  const updateRuleRating = useCallback(
```
AllenBW, line 70: Hard no to this, there is a rule rating component already that does this work

gkarat:
@AllenBW, PTAL 👨🏼‍🔧 

I have tested it locally for RHEL Advisor. Everything works as expected except 403 error when trying to rate a rule (I have attached screenshots). But I suppose it is related to the local environment features, not to the PR's changes.

Screenshot from 2021-08-19 11-48-02
Screenshot from 2021-08-19 11-47-50

gkarat:
For testing with OCP Advisor: well, there is no easy way to do it until we merge it... I will test it out as soon as you push it to fed modules :-)

src/PresentationalComponents/RuleRating/RuleRating.js
```diff
@@ -1,33 +1,25 @@
...
 import React, { useState } from 'react';
 
-import API from '../../Utilities/Api';
 import { Button } from '@patternfly/react-core/dist/js/components/Button/Button';
 import OutlinedThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-down-icon';
 import OutlinedThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/outlined-thumbs-up-icon';
 import PropTypes from 'prop-types';
 import ThumbsDownIcon from '@patternfly/react-icons/dist/js/icons/thumbs-down-icon';
 import ThumbsUpIcon from '@patternfly/react-icons/dist/js/icons/thumbs-up-icon';
-import { injectIntl } from 'react-intl';
```
AllenBW, line 14: Swap this for use intl n instantiate it in the component rather than passing it in

AllenBW (changes requested):
I get 403 on rule rating as well, gotta fix that before this can be merged ```You don't have permission to access "http://console.redhat.com/api/insights/v1/rating/" on this server.```

AllenBW (approved):
💯 🏅 🥅 :godmode: 

this stuff is da bomb, way to fix @gkarat <3

PR #862, opened 2021-08-17
updates to help keep data.json inline with translation files

ddonahue007:
Addition of createDataJson script will help keep data.json file updated with translation files.


AllenBW (approved):
@ddonahue007 THIS IS AWESOME!! Thanks for dropping in 🤗  cleaned up the lock file, locked down the package deps added, reordered the scripts, feel real good about this 😍

PR #863, opened 2021-08-20
Allows for xml to be rendered in report details

AllenBW:
fixes https://issues.redhat.com/browse/ADVISOR-1936

#### because this is fixed, you can see the xml in step 3 of the resolve
Screen Shot 2021-08-20 at 8 20 10 AM


fpjrh (approved):
Thank you for the quick turnaround on this; you're Mr. Fix-It!

PR #864, opened 2021-08-23
Make riskOfChange optional: rule details component

gkarat:
Right now, there is no resolution (risk of change) value
available in the contents of OCP Insights rules. As RuleDetails is a
shared component, we need to make sure the component does not crash or
show incorrect info when this value is absent.


gkarat:
How it looks in the RHEL Advisor (simulated undefined risk of change values):

Screenshot from 2021-08-23 21-35-46
Screenshot from 2021-08-23 21-35-33

gkarat:
@AllenBW, PTAL ⛷️

AllenBW (approved):
LG2M <3

PR #865, opened 2021-08-23
Hide feedback form if there is no onFeedbackChanged provided

gkarat:
OCP infrastructure is designed in a way that the feedback is given to the set of clusterId + ruleId. There are planned changes that should change the behavior to be clusters-wide. But for now, there will be a need to hide the feedback form for OCP recommendation details page.

RHEL example when the callback is not provided:
Screenshot from 2021-08-23 22-28-18


gkarat:
@AllenBW , PTAL 🙏🏼

AllenBW (approved):
simple enough <3

PR #876, opened 2021-10-14
Fix "view affected" message intl disclosure

gkarat:
The patch changes how the `RuleDetails` component decides whether to render the "View affected items" message (see the screenshot). TBH, even the before/after changes look equal, the older way didn't cause this link to be rendered. 

Tested it locally, and now it works as expected.

Frame 1 (8)


karelhala (approved):

gkarat:
Can't proceed as the build is broken

PR #894, opened 2022-01-25
chore(devtools): Configure proxy for stage env by default

bastilian:


config/dev.webpack.config.js
```diff
@@ -12,23 +16,22 @@ const insightsProxy = {
...
-      target: 'https://console.redhat.com/',
-      secure: true,
-      changeOrigin: true,
-      autoRewrite: true,
-      ws: true,
-      onProxyReq: function (request) {
-        request.setHeader('origin', 'https://console.redhat.com/');
-      },
-    },
+    // {
```
bastilian, line 24: I'm not sure what this was meant for, but the frontend works without it and breaks (showing "No access") with it.
If it was meant to be used for something and someone knows, it is still here.

jiridostal:
With 3.2.8 utils package everything works as expected!

bastilian:
@jiridostal Thank you for taking a look! I bumped the packages.

jiridostal (approved):

PR #896, opened 2022-01-26
Advisor-2182: disable autocollapse first row

mkholjuraev:
For more info: https://issues.redhat.com/browse/ADVISOR-2182


johnsonm325 (changes requested):
If I understand this correctly, it seems to work perfectly.

johnsonm325:
@mkholjuraev Looks like just some linting type error stopping you here.

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -185,14 +185,9 @@ const BaseSystemAdvisor = () => {
...
-      const isOpen = rows.length
-        ? entity.length
-          ? entity[0].isOpen
-          : false
-        : key === 0
-        ? true
-        : false;
+
+      const selected = (entity.length && entity[0].selected) || false;
+      const isOpen = (entity.length &&  entity[0].isOpen) || false;
```
bastilian, line 190: Won't `.length` always evaluate to true, since `.filter` will always return an array even if it has no items?

bastilian: Maybe just `entity[0]?.selected` works?

mkholjuraev: I believe **.length** returns **0** value when the array is empty which evaluates to false in condition evaluation. 

`const arra= [];
const secondarr = [1];
console.log(arra.length && 'hello' || 'bye', secondarr.length);`  => will output 'bye', 1

johnsonm325:
https://github.com/RedHatInsights/insights-advisor-frontend/pull/896#discussion_r793955761
@mkholjuraev you're right, `entity.length` would return 0 on an empty array, resulting in a `false` as you expect. That being said, `entity[0]?.selected` would return false on an empty array as well. Either is good code. I leave it to you to decide.

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
bastilian: > I believe .length returns 0 value when the array is empty which evaluates to false in condition evaluation.

Sorry, yeah, @johnsonm325 pointed it out to me as well. Totally forgot about that 0 === false.

mkholjuraev: Actullay if I remove the .length, then the page may break becuase of  an empty entity array (**entity[0].isOpen** and **entity[0].selected**) filtered out on line [#184](https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/SmartComponents/SystemAdvisor/SystemAdvisor.js#L184) We need to check that entity exists. 

Actually I would do this totally different, but this needs to be accross the app and requires a lot of change :)

mkholjuraev: this error is what I would face:
image

bastilian: ACK! Thank you for further investigating and elaborating on it!

mkholjuraev:
there was conflicts, I have resolved. If you agree  I will merge this. Thanks

johnsonm325 (approved):
sorry for taking so long to approve this.

PR #900, opened 2022-01-27
RHIF-15: fix eslint configuration

mkholjuraev:
this PR is to resolve error with the command 'npm run lint:js:fix'. The command is very helpfull to resolve all lint issues at once.

image

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,4 +1,7 @@
-extends: ["@redhat-cloud-services/eslint-config-redhat-cloud-services", "plugin:react-hooks/recommended"]
+root: true
+extends: 
+  - '@redhat-cloud-services/eslint-config-redhat-cloud-services'
+  - 'plugin:react-hooks/recommended'
 globals:
   insights: 'readonly'
   shallow: readonly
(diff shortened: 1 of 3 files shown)


.eslintrc.yml
```diff
@@ -1,6 +1,10 @@
-extends: ["@redhat-cloud-services/eslint-config-redhat-cloud-services", "plugin:react-hooks/recommended"]
+root: true
+extends: 
+  - "@redhat-cloud-services/eslint-config-redhat-cloud-services"
+  - "plugin:react-hooks/recommended"
+  - "prettier"
```
bastilian, line 5: The `prettier` rules should be included via the `eslint-config`[0], or? 


[0] https://github.com/RedHatInsights/frontend-components/blob/master/packages/eslint-config/index.js#L20

mkholjuraev: I applied your suggestion. I hope I understand it correctly.

bastilian: What I meant was, that, since `'@redhat-cloud-services/eslint-config-redhat-cloud-services'` is already included to extend, and that configuration includes `prettier` it should be good to drop those and only have:

```yaml
extends: 
  - "@redhat-cloud-services/eslint-config-redhat-cloud-services"
  - "plugin:react-hooks/recommended"
```

mkholjuraev: ups. fixed

PR #902, opened 2022-02-01
fix: ADVISOR-1903 - Update "Ansible" to "remediation"

bastilian:
This updates all places mentioning "Ansible" to use the new wording.

The changes can be seen on the "Recommendations" page as well as any SystemsAdvisory page and the "Advisor" tab in the Inventory application.


johnsonm325 (approved):
All looks well and seems to work.

PR #903, opened 2022-02-02
fix: ADVISOR-1943 - Update empty states

bastilian:
This updates the empty states mentioned in the Jira.

The Jira also describes the conditions that the empty states should occur in and where.


src/PresentationalComponents/SystemsTable/Components/NoSystemsTable.js
```diff
@@ -0,0 +1,30 @@
...
+  TextVariants,
+  Title,
+  Bullseye,
+} from '@patternfly/react-core';
+
+const NoSystemsTable = () => (
+  <Bullseye>
+    <EmptyState variant={EmptyStateVariant.full}>
+      <Title headingLevel="h2" size="lg" style={{ fontWeight: 'bold' }}>
+        No matching systems found
```
mkholjuraev, line 17: Should we start adding these texts to internationalizatio (**intl.formatMessage()**)

src/Messages.js
```diff
@@ -142,7 +142,13 @@ export default defineMessages({
   ruleIsDisabledBody: {
     id: 'ruleIsDisabledBody',
     description: 'Explaining that the recommendationis disabled',
-    defaultMessage: 'This recommendation has been disabled and has no results.',
+    defaultMessage: 'This recommendation was disabled and has no results.',
+  },
+  ruleIsDisabledBodyWithJustification: {
+    id: 'ruleIsDisabledBodyWithJustification',
+    description: 'Explaining that the recommendationis disabled',
```
mkholjuraev, line 149: Minor: there is a spelling error on line: 144

mkholjuraev (changes requested):
There is one scenario which I think not resolved **Advisor Topics recommendations ->  details page:**

it should show **To find more recommendations,**  but PR shows **To view more recommendations, **

I also did not understand how to test the scenario **2. The suggested rewrite when the recommendation is disabled and the user wrote a reason to a justification note**

src/PresentationalComponents/SystemsTable/Components/NoSystemsTable.js
bastilian: Yeah, let's start adding them to translations. I'll pull out the strings where I find them for this PR.

src/Messages.js
bastilian: Nice catch! Fixed that!

bastilian:
> There is one scenario which I think not resolved Advisor Topics recommendations -> details page:

This should now be corrected. 

>  The suggested rewrite when the recommendation is disabled and the user wrote a reason to a justification note

This should be on the details page of a disabled recommendation. You'll probably need to enter a justification for a recommendation and disable it or use an already disabled one.

mkholjuraev (approved):

PR #908, opened 2022-02-24
[RHIF-20] Add AAP and MSSQL workloads

johnsonm325:
I also added Advisor's first test tests and removed enzyme.


mkholjuraev (approved):
LGTM! I have run the PR, it works as expected. There is one comment that needs a fix.

It is great that Advisor also will have test coverage after this PR. One more minor point, I have noticed that MSQL filter is not covered with the tests.

mkholjuraev (approved):
Tested locally, everything seems good. Thanks.

mkholjuraev (approved):
Looks like my approval got lost after a push.

PR #915, opened 2022-03-03
Add `npm run start:proxy:beta` script

leSamo:
For consistency with other apps


mkholjuraev (approved):
There is one more way: you can achieve the same `"start:proxy:beta": "BETA=true npm run start:proxy"`. 

Otherwise ready to be merged

PR #917, opened 2022-03-07
fix: SystemAdvisor table row count is fixed

mkholjuraev:
Fixes incorrect row count in SystemAdvisor page:

image

### DIFF
--- a/src/SmartComponents/SystemAdvisor/SystemAdvisor.js
+++ b/src/SmartComponents/SystemAdvisor/SystemAdvisor.js
@@ -15,7 +15,7 @@ import {
   TooltipPosition,
 } from '@patternfly/react-core';
 import { IntlProvider, useIntl } from 'react-intl';
-import React, { Fragment, useEffect, useState } from 'react';
+import React, { Fragment, useEffect, useRef, useState } from 'react';
 import {
   SortByDirection,
   Table,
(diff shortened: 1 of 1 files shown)


src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -64,6 +64,7 @@ const BaseSystemAdvisor = () => {
   const [inventoryReportFetchStatus, setInventoryReportFetchStatus] =
     useState('pending');
   const [rows, setRows] = useState([]);
+  const [rowCount, setRowCount] = useState(0);
```
bastilian, line 67: Does it really have to be in a State?

mkholjuraev: hmm.  it makes sense. Do you think useRef might be good option?

mkholjuraev: if I put a simple variable inside component, the variable gets reinitialized every rerender.

bastilian: You could use `useMemo` and make it depend on the `rows`, but generally I feel  just using a simple variable that doesn't require too much computing should be fine.

mkholjuraev: I have changed to useRef.

bastilian: The reason why i'm a bit wary about using a state for this is that it needs to be maintained, so every time the rows change it has to change another state as well.

mkholjuraev: yeap, I have pushed new changes  to implement with useRef. I believe that useRef does not require computation as you said :).

MichalSajdik (approved):

PR #918, opened 2022-03-08
chore: fix error in row count

mkholjuraev:
This PR is intended to fix the incorrect row count after appliying a filter in Advisor System details table.

Before:
image

After:
image


bastilian (approved):
Thank you @mkholjuraev!

PR #921, opened 2022-03-14
fix: reset loading state to active state when dowload fails

mkholjuraev:
We should change current behaviour from showing 'Loading' state when download report fails to active button state. 

Before:
Screenshot from 2022-03-08 13-32-12

After:
image


johnsonm325 (approved):
Looks good. Great job

PR #925, opened 2022-03-17
feat(SystemsTable): ADVISOR-1959 - Use grouped OS selection via helpers

bastilian:
This updates the OS filter to use the helpers provided by the Inventory applications as federated modules.

**Note:** Currently the inventory helpers are only deployed to `/beta`

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -33,11 +33,17 @@ import { useIntl } from 'react-intl';
 import { useLocation } from 'react-router-dom';
 import { usePermissions } from '@redhat-cloud-services/frontend-components-utilities/RBACHook';
 import NoSystemsTable from './Components/NoSystemsTable';
+import { useLoadModule } from '@scalprum/react-core';
 
 const SystemsTable = () => {
   const intl = useIntl();
   const dispatch = useDispatch();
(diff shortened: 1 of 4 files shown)


src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -87,10 +93,18 @@ const SystemsTable = () => {
       values.length > 1 &&
       values.includes('yes') &&
       values.shift();
-    values.length > 0
-      ? setFilters({ ...filters, offset: 0, ...{ [param]: values } })
+    const passValue =
+      param === SFC.rhel_version.urlParam
+        ? Object.values(values || {}).flatMap((majorOsVersion) =>
```
mkholjuraev, line 98: values is treated as an arrayabove (e.g 93th line), while you are treating as an object. I suspect it is intended right?

mkholjuraev (approved):
Tested locally, working like a charm! Thank you.

src/PresentationalComponents/SystemsTable/SystemsTable.js
bastilian: @mkholjuraev yeah, this is intended for the group filters (here the rhel version), since it returns it's value as an object like 
```js
{8.0: {8.0: true, 8.1: true,...},...}
```

But the api/filter functions expect an array.

PR #931, opened 2022-03-24
Changed the Pathway coloumn to say name instead of pathway.

adonispuente:


leSamo (changes requested):
This works well, but working with translations is bit tricky.

We like to keep the `id` of the translation the same as its key, so it does not cause confusion. So I suggest renaming the key and id of this translation to something like this (because `Name` id is already being used)
```
pathwaysName: {
    id: 'pathwaysName',
    description: 'Column title for pathway name in pathways table',
    defaultMessage: 'Name',
  },
```

and then finding the occurence of this translation in code (`PathwaysTable.js`) and replacing
`title: intl.formatMessage(messages.pathway),`
with
`title: intl.formatMessage(messages.pathwaysName),`

You have edited the master `Messages.js` file, but after that you should run `npm run translations` to regenerate and then commit master json translations file `translations.json` and english json translations `data.json`.

But have no worries, these are just conventions and you had no way to know these beforehand.

Let me know if you have any other questions :grinning:

leSamo (approved):
You nailed it :grinning:

PR #932, opened 2022-03-24
Added margin to the incident label

adonispuente:
This ticket is regarding ADVISOR-2142. There is a class , adv-c-label-incident , that is only assigned to incident labels. It was never given any CSS. Im not sure what the standard is or if we have one in particular for spacing, but I added 5px of margin to at least eliminate the spacing problem.


bastilian (approved):

bastilian:

PR #933, opened 2022-03-25
chore: put app tour button under a feature flag

mkholjuraev:
In order to enable/disable the feature flag, you would need to visit: https://insights-stage.unleash.devshift.net/#/features

When the feature is disabled: 
image

When the feature is enabled:
image


src/SmartComponents/Recs/List.js
```diff
@@ -56,6 +57,12 @@ const List = () => {
     history.push(tab === 1 ? '/recommendations/pathways' : '/recommendations');
   };
 
+  const { flagsReady } = useFlagsStatus();
+
+  const isAppTourEnabled =
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    (flagsReady && useFlag(featureFlags.appTour)) || false;
```
bastilian, line 64: tse, tse, tse. :D Why not pack the `useFlagStatus` and `useFlag` into a hook like `useFeatureFlag('appTour')` and use `useState` for the flagReady?

leSamo: Also, is there a benefit to using && and || instead of using ternary operator (? and :)? You could get rid of some parentheses.

mkholjuraev: @bastilian wondering how I can avoid conditional useFlag hook call by using suggested hook?

mkholjuraev: I need to call only if flags are ready.

mkholjuraev: @leSamo I am too much used to logical operators. I will update this to use ternary operators.

mkholjuraev: Never mind, looks like useFlag hook handles non-existing flags nicely. It returns false. It means even if the flags are not fetched from backend, it will not break the page.

PR #935, opened 2022-03-25
Changed the description to be a static text

adonispuente:
As per ticket ADVISOR-2129 , the description no longer matches the title, but instead shows a static test.


leSamo (approved):
Look good :grinning:

PR #941, opened 2022-03-30
ADVISOR-2127

adonispuente:
The ticket description is
'Always keep the reset filters text and default filter available. Right now, when we cancel the filter, it all disappears. AND when removing the filter, it disables the export functionality which is a bug.'

With the changes I made the 'Reset Filter' button is always present, and the export functionality is always available.
To see the changes
Go to Advisor,
Click recommendations,
Delete the two default filters 'systems impacted' , and 'status enabled'
You'll see that you can still select the export button
And you'll see that now the 'Reset Filter' button is still present, and onClick the filters are restored.

### DIFF
--- a/src/SmartComponents/Recs/Details.js
+++ b/src/SmartComponents/Recs/Details.js
@@ -34,7 +34,6 @@ import Failed from '../../PresentationalComponents/Loading/Failed';
 import { Flex } from '@patternfly/react-core/dist/js/layouts/Flex/Flex';
 import { FlexItem } from '@patternfly/react-core/dist/js/layouts/Flex/FlexItem';
 import Inventory from '../../PresentationalComponents/Inventory/Inventory';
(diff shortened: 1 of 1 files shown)


leSamo (changes requested):
As you removed the `isDisabled` prop completely this would break granular RBAC; you see users can have permissions (granted to them by their organization's administrator) to have access to various functionalities of an app. In Advisor they can for example be granted a permission to export Pathways to CSV or JSON. If they don't have the permission the export button should be hidden or disabled, according to UXD standards it should be hidden, but in Advisor it was disabled, but whatever. 

Now by deleting 
```
isDisabled: !permsExport
```
export button will always be enabled, even when user does not have permission to export, which is incorrect. If you need some more explanation on how RBAC works or you want instructions on how to change permissions for testing let me know.

MichalSajdik (dismissed):
LGTM, but update branch before merge

PR #944, opened 2022-03-30
ADVISOR 2384

adonispuente:
When pressing the 'Reset Filter' button on a pathways details page, the filter would reset the pathway to null, and then it would just show all recommendations regardless if they were related to the pathway.

To see changes,
Go to advisor, navigate to pathways table , and select a pathway. The default filters are currently 'systems impacted' and ' status enabled'. Before the change, if you pressed reset filter, it would show all recommendations, now it should still show recommendations for specific pathway.


leSamo (approved):

leSamo (changes requested):
This fix does work, but what I also noticed is that after you change filters `pathway=update-kernel-boot-options` is added as an URL param; we don't need to have pathway param in URL params as it's part of the path (`/advisor/recommendations/pathways/update-kernel-boot-options`), it'd nice to remove it from URL params

https://stage.foo.redhat.com:1337/beta/insights/advisor/recommendations/pathways/update-kernel-boot-options?pathway=update-kernel-boot-options&limit=20&offset=0&sort=category#SIDs=&tags=

adonispuente:
> This fix does work, but what I also noticed is that after you change filters `pathway=update-kernel-boot-options` is added as an URL param; we don't need to have pathway param in URL params as it's part of the path (`/advisor/recommendations/pathways/update-kernel-boot-options`), it'd nice to remove it from URL params
> 
> https://stage.foo.redhat.com:1337/beta/insights/advisor/recommendations/pathways/update-kernel-boot-options?pathway=update-kernel-boot-options&limit=20&offset=0&sort=category#SIDs=&tags=

When I bring down master and navigate to the same page, the URL im getting is https://stage.foo.redhat.com:1337/beta/insights/advisor/recommendations/pathways/update-kernel-boot-options?pathway=update-kernel-boot-options&limit=20&offset=0&sort=category&impacting=true , which has the pathway param in there. I believe the reasoning for it is because the filter builds itself off of the url params. 

What steps are you taking in order to get that URL? I cant seem to produce it on my end.

leSamo:
@adonispuente you are reproducing it :grinning: you see in your URL there is both `/pathways/update-kernel-boot-options` path and `pathway=update-kernel-boot-options` URL param, we have duplicate information about what pathway this is in the URL, let's remove it from URL param, so it's there only once

adonispuente:
@leSamo But wouldnt the filter stop working properly if we remove it? The filter is using paramParser() to dictate what to load. I thought that was why it was in the url twice

leSamo:
Well yes, it may require a small bit of refactoring; what params we send to the backend is not directly read from URL params and sent to backend; editing filters causes data request to be sent with new params and setFilters() to be called which in turn calls urlBuilder() which edits URL params. We can just tell urlBuilder that we don't want to show pathway param in URL, this will have no effect on how data request is called.

adonispuente:
I see what youre saying. In this most recent change, I originally was filtering out the pathway tag within the map method, but for some reason it only worked most of the time. I ended up just deleting the pathway key for the url all together.

PR #951, opened 2022-04-14
ADVISOR-2127

adonispuente:
The ticket description is
'Always keep the reset filters text and default filter available. Right now, when we cancel the filter, it all disappears. AND when removing the filter, it disables the export functionality which is a bug.'

With the changes I made the 'Reset Filter' button is always present, and the export functionality is always available.
To see the changes
Go to Advisor,
Click recommendations,
Delete the two default filters 'systems impacted' , and 'status enabled'
You'll see that you can still select the export button
And you'll see that now the 'Reset Filter' button is still present, and onClick the filters are restored.


MichalSajdik (approved):
LGTM 
Link to previous PR: https://github.com/RedHatInsights/insights-advisor-frontend/pull/941

PR #961, opened 2022-04-21
ADVISOR-2000, bulkselector displays correct sign when item selected

adonispuente:
To view, run the app (must be in beta)

- Click recommendations in the left menu, 
- Selected pathways tab
- Select a pathway
- Click systems tab
- Select a system
Where before the bulkselector would remain empty besides a '1 selected', it now displays a ' -  '

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -429,6 +429,8 @@ const Inventory = ({
               ? 1
               : selected.length === filters.limit
               ? null
+              : selected.length > 0
+              ? null
               : 0,
           onSelect: () => {
             selected.length > 0 ? onSelectRows(-1, false) : bulkSelectfn();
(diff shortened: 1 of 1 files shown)


mkholjuraev (changes requested):
It looks like marking systems selected got affected: 
1. When I hit 'Select all' button, the number shown on bulk select is different that total number. (Maybe there is a reason that I am not aware).
2. if I move to the next page after all systems are selected, all selected items gets lost.

https://user-images.githubusercontent.com/59481011/164676517-92b686ac-a3bf-4801-bd1f-950267f73f42.mp4


Also that  part of code to mark as checked looks very hard to read with a lot of conditions. I believe that it can be simplified a lot.

adonispuente:
I just looked at the master branch and it looks like you might have actually discovered two other bugs. Im getting the same results on the master branch regarding the select all issue and the systems clearing when going to a new page. I'll keep investing in the mean time.

mkholjuraev (approved):
LGTM! @adonispuente thanks.

PR #962, opened 2022-04-22
ADVISOR-2462, bulk selector dropdown is improved.

adonispuente:
Bulk selector dropdown now shoes the correct 'select page' option number, and there is no empty space in the dropdown before changing the items displayed per page. 
To view changes 
- Run the app
- Click recomendations tab in the left menu
- Navigate to pathways
- Select a pathway
- select the systems tab
- open the bulk selector
Here you will see that where the empty space was, there now displays 'Select Page (20 items)'

- Edit the items show per page by clicking the pagination
- select something like 50 per page
In the bulk select, the 'select page' option will be updated with the current items per page

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -374,16 +374,12 @@ const Inventory = ({
               },
             },
             {
-              ...(entities?.rows?.length > filters.limit
-                ? {
-                    title: intl.formatMessage(messages.selectPage, {
-                      items: filters.limit,
-                    }),
-                    onClick: () => {
-                      onSelectRows(0, true);
-                    },
-                  }
-                : {}),
+              title: intl.formatMessage(messages.selectPage, {
+                items: entities?.rows?.length,
+              }),
+              onClick: () => {
+                onSelectRows(0, true);
+              },
             },
             {
               ...(entities?.rows?.length > 0


leSamo (changes requested):
This works well, I just found a small flaw - when there are less items than `page_size`, for example 6 items with page_size 20, select page still has `page_size` in parentheses, it should only be 6 here
select

adonispuente:
ahhh I didnt think of that edge case, these changes should solve that @leSamo , thank you!

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -374,16 +374,15 @@ const Inventory = ({
...
-                    onClick: () => {
-                      onSelectRows(0, true);
-                    },
-                  }
-                : {}),
+              title: intl.formatMessage(messages.selectPage, {
+                items:
+                  entities?.rows?.length > 9
+                    ? entities.perPage
+                    : entities?.rows?.length,
```
leSamo, line 381: Imagine you'd have `per_page `set to 20 and 11 items on a page -- then this condition would be true, because 11 > 9 and you'd show that selecting page would select 20 items, which is not correct, as there are 11, aren't there?

Now from looking at this code I don't know if `entities?.rows?.length` is paginated (meaning it contains all items but at most page_size) or not paginated (contains items from all pages) -- so providing solution for both cases:

If it is the former then you could just pass
`items: entities?.rows?.length`
because this number is number of items, but at most how many fit on the page

If it's the latter then it would be fixed if you took `Math.min` of those two values right? If there are 100 items and `page_size` is 50 then minimum is 50 -- correct. If there are 15 items and `per_page` is 50, then min is 15 -- correct.

adonispuente: I see what youre saying, I had figured since to code is read in cascading order the end result is what the desired result is, but the way youre looking at it is much cleaner. I'll make the necessary edits

leSamo (approved):
This solves the problem described in the ticket.

I also noticed some more problems, I don't know if they have a ticket
- Select page (X), if there are less than page size items, X should be item count not page size
- Select page, navigate to other page, all items get deselected, but they should not have had
- If table is in loading state the select all checkbox is enabled, but should not be
- If table is in loading state Select all option is missing inside the dropdown

PR #963, opened 2022-04-25
ADVISOR-2125: fix: pathway remediation applied to unaffected hosts

MichalSajdik:


src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -68,11 +68,35 @@ const Inventory = ({
...
+          {},
+          {}
+        )
+      )?.data.rules;
+
+      let issues = [];
+      pathways.forEach((rec) => {
+        let filteredSystems = [];
+
+        systems[rec.rule_id].forEach((system) => {
```
mkholjuraev, line 84: I am wondering if systems array has always rule with associated rule_id, if it is not there might be an edge case. It will complain about undefined index/value.

mkholjuraev (commented):
I am not able to run this PR locally due to VPN issues. The code looks clean. One more eye would be extra benefit

src/PresentationalComponents/Inventory/Inventory.js
MichalSajdik: so just to be save, I should change it to `systems[rec.rule_id]?.forEach((system)` ? But it should always contain `rule_id`

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -68,11 +68,35 @@ const Inventory = ({
...
+
+      const systems = (
+        await Get(
+          `${BASE_URL}/pathway/${encodeURI(pathway.slug)}/reports/`,
+          {},
+          {}
+        )
+      )?.data.rules;
+
+      let issues = [];
```
bastilian, line 80: This could be constructed using `.find`, or?

MichalSajdik: Not sure how, can you make code proposal?

bastilian: @MichalSajdik I may have meant `.filter` or a `.map`, to avoid the `.forEach`s. Though, I'm not sure if it would make this better in any aspect.

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -68,11 +68,35 @@ const Inventory = ({
...
+      pathways.forEach((rec) => {
+        let filteredSystems = [];
+
+        systems[rec.rule_id].forEach((system) => {
+          if (selected.includes(system)) {
```
bastilian, line 85: While trying to see if the `issues` construction could be written differently, I noticed the `selected` and how the `remediationDataProvider` is defined and I believe there is a chance that `selected` becomes stale when React does its re-rendering. And it has to redefine the function on each render. Maybe this could be refactored (in another PR) to use `useMemo` and pass in the `selected` instead to not have it as a dependency for the memoized function.

MichalSajdik: Yea that could be definitely done, but as you say in different PR and probably under another task as performance enhancement. But this PR and task corresponds only for a fix of functionality

src/PresentationalComponents/Inventory/Inventory.js
MichalSajdik: Yea, I would be able to somehow rewrite this using these function, but I doubt that it would look better

bastilian (dismissed):
LGTM! Thank you @MichalSajdik!

PR #965, opened 2022-04-27
ADVISOR-2057 Systems are not selected after remediation 

adonispuente:
After remediating systems, the selected systems are no longer selected.
To view changes
- navigate to systems tab in advisor
- select a system with at least 1 tag (ex. $uper_device has 5)
- select any combination of items available in the table
- click remediation button
- follow through remediation process ('create playbook and continue to press next)
- Once you reach the end and items have been added to a playbook, select 'return to application' or the x on the top right of the modal
- here you should see that the previously selected items are no longer selected


johnsonm325 (approved):
This seems good to me

PR #970, opened 2022-04-29
Advisor 2390 - tag filter works properly in advisor systems

adonispuente:
This is a continuation fix overall for global filters not working properly in Advisor systems. The changes in Common/Tables.js and the Tables.fixtures are proven to work  for workloads and are already in qa. However, the changes with this commit are all in SystemsTable.js

Tags were never being sent to the api endpoint, which was the reason they weren't working. With this fix, tags are always sent, and the System table updates properly. To view the changes on this fix,
- Head to advisor systems
- Find a system that has tags associated with it, click one with this logo Screen Shot 2022-04-29 at 5 39 52 PM
- Copy the tags tag source (I dont suggest satelite, it has too many systems associated with it, i suggest using something with 1 system so the changes are more readable, provided tag = HTZocjat)
- Here you'll see that the systems associated with the selected tag, and nothing else, pop up
- If you press 'Reset Filter' , the tag filter is still applied (assuming you havent cleared it out)
- If you clear the tag filter, the table will update properly

You may question why updates where not just made to Common/Tables.js, because that is where workloads and SID global filters are built.
However, the data there is mutated so that it is sent as something like filters[system_profile][not_nilt]: true. If you apply a tag key/value pair (outside of systems_profile) it is mutated as something like filters[tags][contains][null]:[tag_name] and the api cannot read it so it will always return a result of 0. Similar issue is present with or without lines 159-160. It is for this reason that I went with this route. 

I tried to be thorough in my explanation but if there are any questions, please let me know!


src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -275,7 +264,7 @@ const SystemsTable = () => {
             filters,
             workloads,
             SID,
-            // selectedTags,
+            selectedTags,
```
MichalSajdik, line 267: it seems like this value is not used, and should be removed

adonispuente: ahh good catch, didnt notice that

MichalSajdik (approved):

PR #974, opened 2022-05-23
Bulk Selector Refactor for Advisor : ADVISOR-2472

adonispuente:
This change fixes the issues regarding the bulk selector in advisor pathway details page. 
Right now, when trying to view more systems, all selected items are cleared. To replicate

Click recommendations in the left menu,
Selected pathways tab
Select a pathway
Click systems tab
Select a system or two
Press a new page (look at the pagination button, select the '>' button the view items 20-40 as opposed to 1-20. 
Old Behavior: selected is clear

New Behavior: When moving pages , state is carried across. Also if you do a select all and then deselect 1 item, it doesnt switch to (pagesize -1), it instead does (total Systems -1)

With this change state is carried around properly, and ADVISOR-2388 can be worked on more easily as selected items are managed/updated properly.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -5,7 +5,7 @@ import {
   RULES_FETCH_URL,
   SYSTEMS_FETCH_URL,
 } from '../../AppConstants';
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { TableVariant, sortable, wrappable } from '@patternfly/react-table';
 import {
   pruneFilters,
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -348,6 +337,11 @@ const Inventory = ({
             },
             showTags
           );
+          results.results.forEach((row) => {
```
bastilian, line 340: This should rather be a `map` and not mutate the `results`. Like:
```js
const selectedRows = results.results.map((row) => ({ ...row, selected: selected.includes(row.id) }));
....
```

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -424,19 +418,19 @@ const Inventory = ({
                               { name: filters.name }
                             )
                           )?.data?.host_ids;
-
                       console.error(allSystems);
-                      setSelected(allSystems);
-                      bulkSelectfn();
+                      entities.rows.forEach((item) => {
```
bastilian, line 422: same as above with `map`

PR #976, opened 2022-05-27
ADVISOR-2472 : bulk Selector refactor

adonispuente:
This change fixes the issues regarding the bulk selector in advisor pathway details page.
Right now, when trying to view more systems, all selected items are cleared. To replicate

Click recommendations in the left menu,
Selected pathways tab
Select a pathway
Click systems tab
Select a system or two
Press a new page (look at the pagination button, select the '>' button the view items 20-40 as opposed to 1-20.
Old Behavior: selected is clear

New Behavior: When moving pages , state is carried across. Also if you do a select all and then deselect 1 item, it doesnt switch to (pagesize -1), it instead does (total Systems -1)

With this change state is carried around properly, and ADVISOR-2388 can be worked on more easily as selected items are managed/updated properly.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -29,6 +29,7 @@ import { systemReducer } from '../../Store/AppReducer';
 import { updateReducers } from '../../Store';
 import { useIntl } from 'react-intl';
 import downloadReport from '../Common/DownloadHelper';
+import useBulkSelect from './Hooks/useBulkSelect';
 
 const Inventory = ({
   tableProps,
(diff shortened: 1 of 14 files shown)


src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -105,13 +152,12 @@ const Inventory = ({
...
           },
         ],
-        systems: selected,
+        systems: selectedIds,
       };
     }
   };
 
   const onRemediationCreated = (result) => {
-    onSelectRows(-1, false);
```
bastilian, line 114: I believe this was meant to clear the selection after remediation. The bulk select hook returns a `selectNone` function, which we could replace this with.

mkholjuraev (approved):

PR #981, opened 2022-06-07
fix(ci): Fix package locks

bastilian:


package.json
```diff
@@ -56,10 +56,11 @@
   "dependencies": {
     "@babel/runtime": "7.15.4",
     "@patternfly/patternfly": "4.183.1",
-    "@patternfly/react-core": "4.198.19",
-    "@patternfly/react-icons": "4.49.19",
-    "@patternfly/react-table": "4.67.19",
-    "@patternfly/react-tokens": "4.50.19",
+    "@patternfly/react-core": "^4.221.3",
+    "@patternfly/react-charts": "6.14.17",
```
bastilian, line 60: We need to pin `react-charts` to get the pdf generator to install.

PR #983, opened 2022-06-09
fix: Don't override onError for IntlProvider wrapper component

gkarat:
If the format.js translations were not found, Advisor would log errors in production. Removing the line sets the behavior to default: https://formatjs.io/docs/react-intl/components/#onerror and will make errors appear only if `NODE_ENV` is not set to `production`.


bastilian (approved):
THANK YOU @gkarat!

bastilian (approved):

PR #984, opened 2022-06-13
Prod beta

adonispuente:


bastilian:
This change is so small. I'd rather vote for addressing this in master first and than deploy normally and let this also go through QA.

PR #986, opened 2022-06-14
fix for advisor tags on systems page

adonispuente:
On page load for Systems tab in advisor, the tags were sent as an empty object and the api was getting angry. The tags were set to null at first. This fixes selectedTags and now you can add/remove tag filters and you can load the systems page in advisor. Currently a problem in stage env.


mkholjuraev (approved):
@adonispuente clean fix!

PR #987, opened 2022-06-16
Bulk Selector Refactor with loading state

adonispuente:
With this PR bulk selector refactor is completed. When pressing the 'Select All' option in the dropdown, it will select all systems associated with the currently set filters. So if you were to filter down to 400 systems (starting with lets say 2,000), when pressing selectAll, it will making paginated calls to the api and select the 400 systems as opposed to the 2,000. This PR also adds a loading state to the bulk Selector, and makes bulkSelector work as UX team intended it too. 

 npm run start:proxy:beta
To test the Select All
- Navigate to advisor recommendations in Beta Env,
- Select a pathway with alot of systems (you can click the systems affected link)
- With the systems tab selected, press select all in the bulk selector and it will select every system and while promise is resolving a spinner will appear - old behavior was it selected 1000

- Then deselect all
- Filter down the total systems (type in 'rhiq' and it should filter out something like 50 systems if testing in Insights-qa account) 
- Press selectAll in bulk Selector
- All filtered down results are selected as opposed to all systems associated with the pathway

Bulk Selector should 
-have 'select page' option, that simply selects every item on the current page if not already selected
-Pressing the square should only deselect or selectAll systems
-if the current page is selected, pressed 'select Page' should not do anything, it is only meant for that 1 action


src/PresentationalComponents/Inventory/Hooks/useBulkSelect/useBulkSelect.js
```diff
@@ -25,6 +20,7 @@ const useBulkSelect = ({
   itemIdsInTable,
   itemIdsOnPage,
   identifier = 'id',
+  loading,
```
mkholjuraev, line 23: I would suggest renaming 'loading' to isLoading.

src/PresentationalComponents/Inventory/Hooks/useBulkSelect/useBulkSelect.js
```diff
@@ -116,7 +108,7 @@ const useBulkSelect = ({
                 : []),
             ],
             checked,
-            onSelect: !isDisabled ? selectPage : undefined,
+            onSelect: selectedIds?.length > 0 ? () => clear() : selectAll,
```
mkholjuraev, line 111: How about just passing:
onSelect: selectedIds?.length > 0 ? () => clear() : selectAll,
```suggestion
            onSelect: selectedIds?.length > 0 ? clear : selectAll,
```

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -52,38 +43,36 @@ const Inventory = ({
...
+  const [fullFilters, setFullFilters] = useState();
   const [total, setTotal] = useState(0);
   const entities = useSelector(({ entities }) => entities || {});
+
   const addNotification = (data) => dispatch(notification(data));
   const [disableRuleModalOpen, setDisableRuleModalOpen] = useState(false);
   const [curPageIds, setCurPageIds] = useState([]);
   const [pathwayRulesList, setPathwayRulesList] = useState();
   const [pathwayReportList, setPathwayReportList] = useState();
+  const [loading, setLoading] = useState();
```
mkholjuraev, line 55: Here as well, I would suggest renaming 'loading' to isLoading.

mkholjuraev (dismissed):
LGTM! works as expected. Left some minor suggestions.

src/PresentationalComponents/Inventory/Inventory.js
adonispuente: I agree, thats a better name for it

mkholjuraev (approved):

PR #991, opened 2022-06-28
fix: ADVISOR-2569 - Handle empty KCS article node_id value

mhuth:
Fix for https://issues.redhat.com/browse/ADVISOR-2569

A rule doesn't always have a KCS article associated with it, in which case node_id will be blank.  This PR handles that situation.


mhuth:
Showing the rule has an empty node_id:
empty_node_id

Before PR:
before

After PR: 
after

mkholjuraev (approved):
tested locally. LGTM! @mhuth thank you!

PR #992, opened 2022-06-29
Rules check for ADVISOR-2388

adonispuente:
Following up upon the pathways check for advisor-2388 PR, This PR makes it possible for the button status to change properly when selecting recommendations. 

To test you will need to run against prod with `CHROME_ENV=prod-stable npm run start:proxy:beta`
-Navigate to pathways
- First select a pathway with manual remediation, go the systems tab, you'll see that the remediation button is disabled.
-Then do the same thing with a pathway with an automatic remediation available, you'll see the remediation button can be enabled when you select

Then navigate to recommendations page, 
select a recommendation with an automatic remediation playbook, you'll see its possible to enable to button
Follow the same steps for a recommendation with a manual playbook, you'll see the button isnt selectable

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -1,6 +1,6 @@
 import './_Inventory.scss';
 
-import { BASE_URL } from '../../AppConstants';
+import { BASE_URL, RULES_FETCH_URL } from '../../AppConstants';
 import React, { useEffect, useState } from 'react';
 import { TableVariant, sortable, wrappable } from '@patternfly/react-table';
 import { pruneFilters, urlBuilder } from '../Common/Tables';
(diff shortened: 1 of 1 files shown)


mkholjuraev (commented):
I maybe wrong, but if I understand correctly pathwayCheck and ruleCheck functions might be optimized to be done once only on component mount instead of calling them everytime selected systems change. 

This can be done by moving calling those functions  into a hook with empty deps. Because those checking functions do not depend on selected systems. checkRemediationButtonStatus` can be triggered as it is now whenever selectedSystems changes. Remediation button will be disabled whenever systems are not selected and there is not a playbook associated. 

This will prevent exhaustive API request .

adonispuente:
@mkholjuraev You're comment helped me make a small change regarding optimization for the ruleCheck function.  
 With the new changes the api request only run on page load, so the concern I think now would just be cleanliness/readability of the code.
 
 I can leave it as it is now, or if its weird to read, I can still follow your suggestion and add this code block along with the required changes to the functions. I just wanted to be just this was what you were talking about
 
```
 useEffect(() => {
    dispatch({
      type: 'SELECT_ENTITIES',
      payload: {
        selected: selectedIds,
      },
    });
    if (pathway) {
      checkRemediationButtonStatus();
    }
  }, [selectedIds]);

  useEffect(() => {
    if (pathway) {
      pathwayCheck();
    } else {
      rulesCheck();
    }
  }, []);
```

adonispuente:
Had to rebase the branch with the most recent push

mkholjuraev (approved):
LGTM! @adonispuente thank you!

PR #994, opened 2022-07-06
ADVISOR-2289 Refresh Bug 

adonispuente:
These changes are to make headway against the refresh bug. It seems to be having a different result on some set ups so I want to test with more people. To test this PR

- run advisor with npm run start:proxy:beta
- Select the systems affected link (or click a pathway, and then hit the systems tab)
- select systems
- select the remediate button
- type something inside the input 'create a playbook'

Monitor to see if the data gets refresh/wiped. Ideally with this PR the data is never wiped and the 'removeQueryResult' redux call is never called

### DIFF
--- a/src/SmartComponents/Recs/Recs.js
+++ b/src/SmartComponents/Recs/Recs.js
@@ -19,54 +19,92 @@ const ClassicRedirect = lazy(() =>
   import(/* webpackChunkName: "ClassicRedirect" */ '../Common/ClassicRedirect')
 );
 
-const suspenseHelper = (component) => (
-  <Suspense fallback={<Loading />}>{component}</Suspense>
+// eslint-disable-next-line react/prop-types
+const SuspenseHelper = ({ children }) => (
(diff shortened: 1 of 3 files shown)


src/SmartComponents/Recs/Recs.js
```diff
@@ -22,6 +22,11 @@ const ClassicRedirect = lazy(() =>
 const suspenseHelper = (component) => (
   <Suspense fallback={<Loading />}>{component}</Suspense>
 );
+
+const SuspenseHelper = ({ children }) => (
```
bastilian, line 26: I'm not sure if this is what makes it behave normally, but either way we could wrap the remaining route components with this as well and remove the `suspenseHelper` function above.

gkarat:
Tested locally with the scenario in description, and there was no issue found by me (data is preserved, and no refresh like in the Jira's video)

bastilian (approved):

PR #1000, opened 2022-07-15
ADVISOR-2546: expanded first item on page load

adonispuente:
This is regarding ADVISOR-2546

1) Go to the recommendation list
2) Click on any recommendation with affected systems
3) Click on any system
4) The first recommendation should be open

### DIFF
--- a/src/SmartComponents/SystemAdvisor/SystemAdvisor.js
+++ b/src/SmartComponents/SystemAdvisor/SystemAdvisor.js
@@ -173,7 +173,8 @@ const BaseSystemAdvisor = () => {
     filters,
     rows,
     searchValue = '',
-    kbaLoading = false
+    kbaLoading = false,
+    isFirstLoad = false
   ) => {
     const builtRows = activeReports.flatMap((value, key) => {
       const rule = value.rule;
@@ -184,7 +185,7 @@ const BaseSystemAdvisor = () => {
 
       const match = rows.find((row) => row?.rule?.rule_id === rule.rule_id);
       const selected = match?.selected;
-      const isOpen = match?.isOpen || false;
+      const isOpen = match?.isOpen || (isFirstLoad && key === 0);
 
       const reportRow = [
         {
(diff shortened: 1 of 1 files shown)


src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -184,7 +184,8 @@ const BaseSystemAdvisor = () => {
 
       const match = rows.find((row) => row?.rule?.rule_id === rule.rule_id);
       const selected = match?.selected;
-      const isOpen = match?.isOpen || false;
+      const isOpen = match?.isOpen || false || key === 0;
```
mkholjuraev, line 187: I think we can remove hard coded `false`
```suggestion
      const isOpen = match?.isOpen || key === 0;
```

adonispuente: When going back to implement this and test, I realized that it was opening the first item every time you selected something. In the next commit Im passing in an argument 'firstLoad' so that this will only happen on first initial load

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -625,6 +634,7 @@ const BaseSystemAdvisor = () => {
             filters,
             rows,
             searchValue,
+            true,
```
adonispuente, line 637: Here im just adding a coma, setting 'firstLoad' as true

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -464,7 +465,15 @@ const BaseSystemAdvisor = () => {
...
       setKbaDetailsData(kbaDetailsFetch);
       setRows(
-        buildRows(reportsData, kbaDetailsFetch, filters, rows, searchValue)
+        buildRows(
+          reportsData,
+          kbaDetailsFetch,
+          filters,
+          rows,
+          searchValue,
+          false,
```
adonispuente, line 474: Currently I dont see exactly what the point of the kbaLoading arg is, its never used in any buildRow calls. I have another PR in the making for an advisor refactor, and will look into this more in that PR.

mkholjuraev (approved):
LGTM! thanks @adonispuente.

PR #1005, opened 2022-07-27
fix: Remove intl on error redirect to console

bastilian:
Removes the redirects to console log for Intl on error, which can lock up the browser.


adonispuente (approved):
\o/ Thank you, LGTM!

PR #1006, opened 2022-07-28
style(ADVISOR-2645) SPUD-UI Changes

adonispuente:
Styling changes from Advisor SPUD meeting. 
Will be looked over by Stefan , once approved, just needs code review

### DIFF
--- a/src/PresentationalComponents/Cards/Pathways.js
+++ b/src/PresentationalComponents/Cards/Pathways.js
@@ -69,7 +69,9 @@ const PathwayCard = (props) => {
       </CardBody>
       <CardBody className="body pf-u-font-size-sm">{description}</CardBody>
       <CardBody className="body pf-u-font-size-sm">
-        {has_incident && <RuleLabels rule={{ tags: 'incident' }} isCompact />}{' '}
+        {has_incident && (
+          <RuleLabels rule={{ tags: 'incident' }} isCompact noMargin={true} />
+        )}{' '}
         {RebootRequired(reboot_required)}
       </CardBody>
       <CardFooter className="footer pf-u-font-size-sm">
@@ -105,7 +107,7 @@ const TotalRisk = (props) => {
     >
       <CardTitle>{intl.formatMessage(messages.totalRiskPathway)}</CardTitle>
       <CardBody className="body">
-        <Grid>
+        <Grid hasGutter>
           <GridItem span={6}>
             <div>
               <Chart
@@ -170,7 +172,7 @@ const TotalRisk = (props) => {
               </Chart>
             </div>
           </GridItem>
-          <GridItem span={6}>
+          <GridItem span={6} className="pf-u-font-size-sm">
             {intl.formatMessage(messages.thisPathway, {
               category: catString(categories),
               systems: impacted_systems_count,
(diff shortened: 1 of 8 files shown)


kuklas:
Thumbs up from UX to these changes!

src/PresentationalComponents/Cards/Pathways.js
```diff
@@ -69,7 +69,9 @@ const PathwayCard = (props) => {
       </CardBody>
       <CardBody className="body pf-u-font-size-sm">{description}</CardBody>
       <CardBody className="body pf-u-font-size-sm">
-        {has_incident && <RuleLabels rule={{ tags: 'incident' }} isCompact />}{' '}
+        {has_incident && (
+          <RuleLabels rule={{ tags: 'incident' }} isCompact noMargin={true} />
```
mkholjuraev, line 73: It seems to me that we can remove ={true} and just pass noMargin.
```suggestion
          <RuleLabels rule={{ tags: 'incident' }} isCompact noMargin />
```

src/PresentationalComponents/Cards/Pathways.js
```diff
@@ -196,34 +198,35 @@ const Resolution = (props) => {
       <div className="flex-coloumn">
         <CardTitle>{intl.formatMessage(messages.resolution)}</CardTitle>
 
-        <div className="flex-row">
-          <div>
+        <div className="flex-row ">
+          <div className="halfWidth">
```
mkholjuraev, line 202: Wondering if we can achieve half width layout with a Grid system instead of plain css? I have not tried myself, if this does not make sense please ignore this :)

src/PresentationalComponents/Labels/RecommendationLevel.js
```diff
@@ -8,8 +8,8 @@ const RecommendationLevel = (props) => {
   const intl = useIntl();
   const { recommendation_level: lvl, isCompact } = props;
 
-  const label = (text, lvl, color, isCompact) => (
-    <Label color={color} isCompact={isCompact}>{`${text} - ${lvl}%`}</Label>
+  const label = (text, lvl, color) => (
+    <Label color={color} isCompact={true}>{`${text} - ${lvl}%`}</Label>
```
mkholjuraev, line 12:
```suggestion
    <Label color={color} isCompact>{`${text} - ${lvl}%`}</Label>
```

src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
```diff
@@ -54,7 +54,10 @@ const PathwaysPanel = () => {
         {isFetching ? (
           <Loading />
         ) : !isError && data.data?.length ? (
-          <Grid hasGutter md={4} sm={12}>
+          <Grid
+            md={data.data?.length === 1 ? 8 : data.data?.length === 2 ? 6 : 4}
```
mkholjuraev, line 58: We can make it more readable in my opinion by adding parenthesis into the first else block.
```suggestion
            md={data.data?.length === 1 ? 8 : (data.data?.length === 2 ? 6 : 4)}
```

mkholjuraev (approved):
LGTM! just minor suggestions. @adonispuente thanks!

src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
adonispuente: Originally I tried this, but the linter gets angry

src/PresentationalComponents/Cards/Pathways.js
adonispuente: Grid comes with some default css, thats needs to be overwritten if I bring it in. Eventually I'd like to change most of this with <Flex> and <FlexItem> Elements, via a function. I had opted for this for now because using the Flex elements makes the file look kind of crazy, and at least right now I think this version is much more readable

mkholjuraev (approved):

PR #1014, opened 2022-08-04
chore(labels) update stragling labels to be compact'

adonispuente:
Doing a final walk through, there were some straggling labels not previously seen in the SPUR review. This updates the remaining labels to be compact.


Fewwy (approved):
Checked the changes locally. LGTM

PR #1016, opened 2022-08-09
Advisor-2673 - Refactor(api): matching GET req params in pathwaysTables with Pathways…

adonispuente:
This makes the order of pathways in pathways panel match the order in pathways table and sorts via recommendation level as opposed to impacted systems. In the attached pictures you can see in the old UI, the ordering in panels doesnt match. With the passed in params, the order now matches.

Old UI: 
Screen Shot 2022-08-09 at 12 35 22 PM

New UI: 
Screen Shot 2022-08-09 at 12 34 37 PM

### DIFF
--- a/src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
+++ b/src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
@@ -17,13 +17,19 @@ import messages from '../../Messages';
 import propTypes from 'prop-types';
 import { useGetPathwaysQuery } from '../../Services/Pathways';
 import { useIntl } from 'react-intl';
+import { useSelector } from 'react-redux';
 
 const PathwaysPanel = () => {
   const intl = useIntl();
+  const { sort, offset } = useSelector(
+    ({ filters: { pathState } }) => pathState
+  );
   const [expanded, setExpanded] = useState(
     JSON.parse(localStorage.getItem('advisor_pathwayspanel_expanded') || 'true')
   );
   const { data, isLoading, isFetching, isError } = useGetPathwaysQuery({
+    sort,
+    offset,
     limit: 3,
   });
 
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
```diff
@@ -17,14 +17,18 @@ import messages from '../../Messages';
...
+import { useSelector } from 'react-redux';
 
 const PathwaysPanel = () => {
   const intl = useIntl();
+  const filters = useSelector(({ filters }) => filters.pathState);
   const [expanded, setExpanded] = useState(
     JSON.parse(localStorage.getItem('advisor_pathwayspanel_expanded') || 'true')
   );
   const { data, isLoading, isFetching, isError } = useGetPathwaysQuery({
+    sort: filters.sort,
```
bastilian, line 29: You could just spread the `filters` const in, or?

adonispuente: We dont want all the same params passed in. Mostly just want how they're sorted to be the same. I had throw in offset just incase we changed it in the future. The limit for them need to be different, but I guess I can manipulate it to only show the first 3 in the array (limit is normally 20) if that seems nicer to just spread in {...filters}

bastilian: maybe rewriting line 24, to:

```js
const { sort, offset } = useSelector(({ filters: { pathState } }) => pathState);
```

then you can write `{ sort, offset, limit: 3 }` here.

adonispuente: hmmm, that does look nicer, ill shoot that up

bastilian (approved):

PR #1017, opened 2022-08-10
fix(ADVISOR-2289):Refresh Bug - adds arg to RTK querys to speed up cache clearing

adonispuente:


mkholjuraev (approved):
LGTM! I did not encounter the issue. Thanks @adonispuente

PR #1018, opened 2022-08-11
fix(ADVISOR-2546) : Adds Url check to isOpen variable and sorts list

adonispuente:
This PR adds a url check to the isOpen variable that controls when the first item in the recommendations list is expanded or not. When coming from systems, the first item will no longer be expanded, but when coming from recommendations, the first item will be expanded. 
To test
1st check:
Navigate to recommendations list via 
Recommendations Tab In Navbar> click recommendation under recs tab > Click on a system > see recommendation list
Desired behavior : first item is expanded, and first item is the recommendation you came from

2nd Check
Systems Tab  in Navbar > select a system > see recommendation list
Desired Behavior : first item is not expanded

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -1,11 +1,12 @@
 import './_Inventory.scss';
 
 import { BASE_URL, RULES_FETCH_URL } from '../../AppConstants';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useCallback } from 'react';
 import { TableVariant, sortable, wrappable } from '@patternfly/react-table';
 import { pruneFilters, urlBuilder } from '../Common/Tables';
 import { useDispatch, useSelector, useStore } from 'react-redux';
 import { getEntities, allCurrentSystemIds } from './helpers';
+import { Link } from 'react-router-dom';
 
 import DisableRule from '../../PresentationalComponents/Modals/DisableRule';
 import { Get } from '../../Utilities/Api';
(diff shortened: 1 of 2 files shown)


src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -176,16 +179,28 @@ const BaseSystemAdvisor = () => {
     kbaLoading = false,
     isFirstLoad = false
   ) => {
-    const builtRows = activeReports.flatMap((value, key) => {
+    let isRecParamPresent =
+      window.location.href.indexOf('recommendations') > -1 ? true : false;
+    let newActiveReportsList = activeReports;
+
+    if (isRecParamPresent) {
+      let activeRec = location[2];
```
bastilian, line 187: Does this work for both beta/non-beta?

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -176,16 +179,28 @@ const BaseSystemAdvisor = () => {
     kbaLoading = false,
     isFirstLoad = false
   ) => {
-    const builtRows = activeReports.flatMap((value, key) => {
+    let isRecParamPresent =
+      window.location.href.indexOf('recommendations') > -1 ? true : false;
```
bastilian, line 183: I kinda feel like this wether or not it is expanded should be solved differently without having to inspect the path.

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
adonispuente: Yeah, this is very similar to how breadcrumbs currently work

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
adonispuente: Did you have something else in mind? At least right now I cant think of an alternative that will work on both envs. If an alternative is created we could also probably change how breadcrumbs work. Why dont you like inspecting the path though? Because of how routes are set up right now, I cant think of a scenario where this wouldnt work properly

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -166,25 +167,43 @@ const BaseSystemAdvisor = () => {
...
     rows,
     searchValue = '',
-    kbaLoading = false
+    kbaLoading = false,
+    isFirstLoad = false
   ) => {
-    const builtRows = activeReports.flatMap((value, key) => {
+    const url = window.location.href;
+    let newActiveReportsList = activeReports;
+    let isRulePresent = url.indexOf('activeRule') ? true : false;
```
bastilian, line 183: `useParams` might be better for this.

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -266,39 +267,53 @@ const Inventory = ({
...
+      let systemProfile = defaultColumns.filter(
+        ({ key }) => key === 'system_profile'
+      );
+      let tags = defaultColumns.filter(({ key }) => key === 'tags');
+
+      displayName = {
+        ...displayName[0],
+        transforms: [sortable, wrappable],
+        props: { isStatic: true },
+        renderFunc: (name, id) => {
```
bastilian, line 287: We might want to only pass in a `renderFunc` if there is a `rule` set, otherwise it might link to somewhere unexpected.

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -166,25 +167,43 @@ const BaseSystemAdvisor = () => {
...
+    if (isRulePresent) {
+      let activeRule = location[2];
```
bastilian, line 185: I think with `useParams` you can use `params.id` or something like that, since the route for this is defined as `"/recommendations/:id/:inventoryId/"`.

adonispuente: useParams returns undefined , restructured or not. Im not sure if this is because its systemAdvisor is going through redux and federated modules, but originally when making this I tried something like 
  ```
const { id } = useParams();
  const { inventoryId } = useParams();
  or even  let params = useParams();
```
Trying again now im getting the same thing. I get undefined every time, that was originally why I opted for using useLocation.

bastilian (approved):

PR #1019, opened 2022-08-17
refactor(Advisor) : Advisor Refactor 2 

adonispuente:
This refactor reduces the size of SystemAdvisor, moves EmptyState.js file up into presentational components and into separate files.


src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -113,90 +118,32 @@ const Inventory = ({
         selected: selectedIds,
       },
     });
-    checkRemediationButtonStatus();
```
bastilian, line 116: Maybe this should/could be a hook, something like `useRemediationButtonStatus`, that includes the  `isRemediationButtonDisabled` state and returns it. Maybe the same can be done with the `pathwayCheck` below.

src/SmartComponents/Recs/helpers.js
```diff
@@ -72,3 +88,72 @@ export const bulkHostActions = async ({
     });
   }
 };
+
+const DownloadExecReport = ({ isDisabled }) => {
```
bastilian, line 92: Since this is a component, it could also live in its own file, maybe under `./components`.

src/PresentationalComponents/Inventory/Inventory.js
adonispuente: For pathway/rule Check, since they only run once on page load, what would be the benefit of changing those to hooks?

src/PresentationalComponents/Inventory/helpers.js
```diff
@@ -152,3 +157,100 @@ export const allCurrentSystemIds =
...
+};
+
+export const useRemediationButtonStatus = (
+  pathwayReportList,
+  selectedIds,
+  pathway,
+  pathwayRulesList,
+  rulesPlaybookCount
+) => {
+  let playbookFound = false;
```
bastilian, line 217: This looks like it should be a state.

src/PresentationalComponents/Inventory/helpers.js
```diff
@@ -152,3 +157,100 @@ export const allCurrentSystemIds =
...
+  useEffect(() => {
+    if (selectedIds?.length <= 0 || selectedIds === undefined) {
+      setIsRemediationButtonDisabled(true);
+    } else if (pathway) {
+      for (let i = 0; i < selectedIds?.length; i++) {
+        let system = selectedIds[i];
+        if (playbookFound) {
+          break;
+        }
+        ruleKeys.forEach((rule) => {
```
bastilian, line 231: `forEach` doesn't seem right here. This could probably be done with filter/any, or?

adonispuente: We want to check against each ruleKey. ruleKeys are used to see if the report list[ruleKey] has the system we are currently on. If reportList[ruleKey] doesn't have it, we need to move on to the next one. Typically there arent a lot of rule keys, but each needs to be looked through. using find() wouldnt allow for the checking of each rulekey against pathwayReportList. 
Thats why the find() is currently used on line 236. At least I dont yet see how to implement find() there

bastilian: I've been looking into this function, and maybe let's go over it together. I think we can make it way more efficient and clean.

PR #1020, opened 2022-08-22
 Refactor(advisor): System Advisor sorts by total risk on first load when coming from systems route

adonispuente:
Per pm, this introduces sorting by total_risk when coming from any route other than recommendations tab. 
These are the two routes to verify 

- Advisor recommendations -> recom details -> Affected system: first in the table should be the recommendation in question, expanded.
- Advisor systems -> system details: all recommendations are collapsed, sorted by total risk.


mkholjuraev (approved):
LGTM! working as described! @adonispuente thanks

PR #1022, opened 2022-08-24
Refactor(ADVISOR-2673): PathwaysPanel is sorted by rec lvl, independent of PathwaysTable 

adonispuente:
This refactors PathwayPanel so that it is independent of the sort on pathwaysTable. It will always be sorted via recommendation level.

### DIFF
--- a/src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
+++ b/src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
@@ -21,14 +21,12 @@ import { useSelector } from 'react-redux';
 
 const PathwaysPanel = () => {
   const intl = useIntl();
-  const { sort, offset } = useSelector(
-    ({ filters: { pathState } }) => pathState
-  );
+  const { offset } = useSelector(({ filters: { pathState } }) => pathState);
   const [expanded, setExpanded] = useState(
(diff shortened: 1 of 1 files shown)


src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
```diff
@@ -21,14 +21,12 @@ import { useSelector } from 'react-redux';
...
-  const { sort, offset } = useSelector(
-    ({ filters: { pathState } }) => pathState
-  );
+  const { offset } = useSelector(({ filters: { pathState } }) => pathState);
   const [expanded, setExpanded] = useState(
     JSON.parse(localStorage.getItem('advisor_pathwayspanel_expanded') || 'true')
   );
   const { data, isLoading, isFetching, isError } = useGetPathwaysQuery({
-    sort,
+    sort: '-recommendation_level',
```
bastilian, line 29: Wouldn't this break allowing users to sort by other columns? Like systems or name?

adonispuente: The desired behavior is that pathway panels is static. We dont want it to be sorted any other way. Also as it is now, pathway panels reload every time pathway tables is sorted because the 'sort' filter in redux changes. @bastilian

bastilian (approved):

PR #1024, opened 2022-09-02
Only apply filters on page load if tab is active

leSamo:
[ADVISOR-2735](https://issues.redhat.com/browse/ADVISOR-2735)


adonispuente (approved):
PR works as intended, LGTM!

PR #1025, opened 2022-09-02
Redo README.md to be up to date

leSamo:
Remove outdated information (running with Insights proxy, old environments, running with custom API) and replace it wit useful information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@
 2. Run development server with `npm run start:proxy:beta`
 3. Local version of the app will be available at https://stage.foo.redhat.com:1337/beta/insights/advisor/
 
+In case you want to use the stable environment instead of beta you can run the app with `npm run start:proxy` and access it from https://stage.foo.redhat.com:1337/insights/advisor/. Usually there is no difference between these two environments unless there is a large feature in progress which is hidden behind `isBeta` flag.
+
 ## Testing
 Travis is used to test the build for this code.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,71 +1,47 @@
...
-- prod-stable, prod-beta
-- ci-stable, ci-beta
-- qa-stable, qa-beta
-
-A push or merge to master will automatically release to ci-beta and qa-beta
-The same will happen with action against master-stable, it will automatically release to ci-stable and qa-stable
-The prod-beta and prod-stable environments are updated by a deliberate push, (to each branch)
+## Running locally
+1. Install dependencies with `npm install`
+2. Run development server with `npm run start:proxy:beta`
```
adonispuente, line 12: At the moment theres really no difference between stage and stage-beta with the pathways release, but I still think its a good idea to include that you have the option to run against stage env via `npm run start:proxy` and then specifying that it would be found on its respective env , https://stage.foo.redhat.com:1337/insights/advisor/.

Now that pathways is released it might even be better to update the bash script so that pushing to stage-beta env also pushes to stage

leSamo: Okay, sounds reasonable, I added that, I think it might fall out of fashion soon with more usage of Unleash, but we can remove it afterwards, thanks for review.

adonispuente (approved):
Great PR, LGTM!

PR #1026, opened 2022-09-02
feat(ADVISOR-2125): Add new ErrorState from FEC [Do Not Merge]

adonispuente:
This PR implements the Error State from FEC, and updates the error state with a new message in pathwayPanels. 

The break api request and see the error states, navigate to these files

/SmartComponents/Topics/Topics.js
/Services/Pathways.js
/Services/Recs.js


src/PresentationalComponents/PathwaysTable/PathwaysTable.js
```diff
@@ -380,9 +380,9 @@ const PathwaysTable = ({ isTabActive }) => {
       {isFetching ? (
         <Loading />
       ) : isError ? (
-        <Failed
-          message={intl.formatMessage(messages.rulesTableFetchRulesError)}
-        />
+        <Table>
```
mkholjuraev, line 383: While running the PR locally, I have seen that the table toolbar exists when there is an error. Do we want them to be displayed? Enabled filters and pagination might bring a negative user experience on a broken page in my opinion.

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -765,9 +765,9 @@ const RulesTable = ({ isTabActive }) => {
       {isFetching ? (
         <Loading />
       ) : isError ? (
-        <Failed
-          message={intl.formatMessage(messages.rulesTableFetchRulesError)}
-        />
+        <Table>
```
mkholjuraev, line 768: Here as well.

src/PresentationalComponents/PathwaysTable/PathwaysTable.js
adonispuente: Actually. I thought that would be a good thing in those instances where a table has the toolbar. I thought that if for whatever reason if a user breaks the table filtering down (for some reason an api fails) , they have the option of pressing 'reset filters' in order to refresh the table with refreshing the page. However, if what youre recommending is the case for the other apps, then it does make more sense to be consistent with the rest of the apps. Is this the case?

mkholjuraev: Yes, I vaguely remember the UX asked to hide the primary toolbar. Thus, at least Vuln and Patch hide the primary toolbar. I think that discussing this with the Advisor UX would be nice.

adonispuente:
@mkholjuraev I've gotten confirmation from Katie saying the toolbar is to be present. She's writing up the guidelines today

mkholjuraev (approved):
LGTM, I did not run the PR this time though as I run it previously.

PR #1029, opened 2022-09-12
fix(SystemTable): update the OS column sort key

mkholjuraev:
This updates the os column sort key after it has been changed on the Inventory side ([commit](https://github.com/RedHatInsights/insights-inventory-frontend/commit/ef34976e44fa0e9a7231579f981d537dfaab408d)).

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -273,7 +273,7 @@ const SystemsTable = () => {
           } = config;
           const sort = `${orderDirection === 'ASC' ? '' : '-'}${
             (orderBy === 'updated' && 'last_seen') ||
-            (orderBy === 'system_profile' && 'rhel_version') ||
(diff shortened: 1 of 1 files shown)


adonispuente:
This PR updates the sort key, which allows it not to break all together, but the sorting does not work.
When click the OS column, You get all the unavailable ones, but then you cannot sort by OS afterwards. 

Attached is the behavior im observing:


https://user-images.githubusercontent.com/60629070/189994824-034ab874-d88a-47c9-a006-431e1142ef8f.mp4

mkholjuraev:
@adonispuente I forgot to mention that you need to run this PR against only stage/beta as the inventory update exists on that env only currently. I tried running the PR, it is working as expected.

https://user-images.githubusercontent.com/59481011/190356131-9480e93b-d04f-4463-87d8-beed8297a9b9.mp4

I will rebase this PR with the master to clean up the commit history.

adonispuente (approved):
Running on beta env I see that this does work. We just have to be mindful to make sure this commit is only pushed when it goes to stage and prod in inventory.

PR #1031, opened 2022-09-26
Do not check SYSTEM_FILTER_CATEGORIES while generating OS filter chips

gkarat:
Fixes https://issues.redhat.com/browse/ADVISOR-2780.

This removes constants for OS filter, and the chips are no longer checked with the SYSTEM_FILTER_CATEGORIES config file. Previously, it led to the bug with OS filter chips missing "RHEL" for some of versions not listed in category.values at SYSTEM_FILTER_CATEGORIES (for example, 9.1).


src/PresentationalComponents/Common/Tables.js
```diff
@@ -86,21 +86,31 @@ export const pruneFilters = (localFilters, filterCategories) => {
   const prunedFilters = Object.entries(localFilters);
   return prunedFilters.length > 0
     ? prunedFilters.reduce((arr, item) => {
+        console.log('@@@', arr, item);
```
adonispuente, line 89: is this console log intentional?

gkarat: oopsie. removed

mkholjuraev (approved):
LGTM!

PR #1033, opened 2022-10-05
fix(ADVISOR-2791): Fixes sorting in Inventory table

adonispuente:
This fixes sorting in the Inventory table, while also disabling sorting for columns that are not supported by the backend.
Tickets for columns [Impacted_date, rhel_version(OS)] have been made and the backend team is aware. Once backend supports these, we can enable sorting on the frontend. 

To test navigate to recommendations - select a recommendation > try sorting on the table there with the available options.

P.S I will squash the commits. I didn't mean to grab all of these.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -95,7 +95,8 @@ const Inventory = ({
     selectedIds,
     setFullFilters,
     fullFilters,
-    rule
+    rule,
+    setFilters
   );
 
   const grabPageIds = () => {
@@ -317,6 +318,7 @@ const Inventory = ({
 
       systemProfile = {
         ...systemProfile[0],
+        sortKey: 'rhel_version',
         transforms: [wrappable],
       };
 
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -318,8 +318,8 @@ const Inventory = ({
 
       systemProfile = {
         ...systemProfile[0],
-        sortKey: 'rhel_version',
         transforms: [wrappable],
+        props: { isStatic: true },
       };
 
       tags = {
--- a/src/PresentationalComponents/Inventory/helpers.js
+++ b/src/PresentationalComponents/Inventory/helpers.js
@@ -11,7 +11,6 @@ export const paginatedRequestHelper = async ({
   filters,
   workloads,
   SID,
-  sort,
   pathway,
   rule,
 }) => {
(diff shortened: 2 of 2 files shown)


mhuth:
@adonispuente the /api/insights/v1/rule/<rule>/systems_detail endpoint does support sorting by impacted_date.  That's the endpoint that's used by the Recommendations system table.  However the main systems table, which is the one you are probably thinking of, doesn't provide an impacted date though.  An impacted_date is only for a system affected by a particular rule, hence why the recommendations system table has it and the general systems table doesn't.  Does that information help?

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -319,14 +320,16 @@ const Inventory = ({
...
-        transforms: [sortable, wrappable],
-        props: { width: 15 },
+        transforms: [wrappable],
+        props: { width: 15, isStatic: true },
         renderFunc: lastSeenColumn.renderFunc,
       };
 
       systemProfile = {
         ...systemProfile[0],
+        sortKey: 'rhel_version',
```
adonispuente, line 330: this is more of a placeholder for now, ATM I've made this unsortable as it is not supported by the backend. Eventually this sortKey might be system_porfile or something but its TBA.

adonispuente:
@mhuth with the updates on the tickets I think we're on the same page now. This PR now just addressed the issue you found here https://issues.redhat.com/browse/ADVISOR-2791 

I've disabled OS sorting. 
In the /systems_details endpoint and /systems endpoint , system_profile is not supported at all

mhuth:
@adonispuente I've created https://gitlab.cee.redhat.com/insights-platform/advisor-backend/-/merge_requests/436 to allow sorting by system_profile (the rhel version) in the backend

src/PresentationalComponents/Inventory/helpers.js
```diff
@@ -35,7 +35,6 @@ export const paginatedRequestHelper = async ({
 
   workloads &&
     (options = { ...options, ...workloadQueryBuilder(workloads, SID) });
-
```
mkholjuraev, line 39: Is this intentional?

mkholjuraev (approved):
Looks goood to me. Works as described. There is one thing that caught my attention, but this does not deal with this PR. It will be better if we try to fix this in later PRs. When you change the sort multiple times, there is one redundant request with the old sorting param:

https://user-images.githubusercontent.com/59481011/195019596-9c5e4df5-4262-4bb4-be1d-70ea0ff7e54c.mp4

adonispuente:
@mhuth I agree with making it unsortable for now. However I see that there is an issue if the sort key differs from the key. I originally tried to make the sortKey for the OS 'rhel_version', but then it would break the sorting (you could not click it more than once, though the request that goes through works fine). Im sure that table in the backend communicates with other tables so I dont think itd be a simple renaming 

@mkholjuraev thank you for the review, that removable or space was not intentional. I will repush now

mhuth (approved):
Looks good to me!  We'll get OS sorting working some other time but there seems to be some work to do on the backend.

PR #1037, opened 2022-10-19
refactor(AppReducer): ESSNTL-3505 - Remove unneeded last column mangling

bastilian:
With https://github.com/RedHatInsights/insights-inventory-frontend/pull/1668 this would cause an issue. However, after investigating it seems that this does nothing, and should be safe to be removed.

Left: Local-Dev (this PR) Right: Stage

Screenshot 2022-10-19 at 13 23 11


adonispuente (approved):
I couldnt find anything having been affected by this, LGTM!

PR #1039, opened 2022-10-25
Run cypress in PRs, and use npm script instead

gkarat:
This PR modifies the cypress workflow config so that the tests are also run in pull requests. Also, it makes the check use the cypress binaries right from `node_modules` instead of using a GH action (which led to flaky behavior sometimes).


bastilian (approved):
LGTM! Thank you @gkarat!

PR #1040, opened 2022-11-01
fix(ADVISOR-2830): Adds workload,SID, and GlobalTags to PathwayPanels

adonispuente:
Global filters and such were not added to pathway panels GET request. 
This causes situations like the one described in the ticket https://issues.redhat.com/browse/ADVISOR-2830 

It could also cause an issue that the pathway table show no pathways (because you add some global filter) but the pathway panels do.

To test, 
Navigate the /recommendations and select the pathways tab
Select global filters and verify that the pathway panel now also reloads when changing global filters. 
The results **should** be identical and in the same order **unless** you altered the sort/added additional filters to the table

### DIFF
--- a/src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
+++ b/src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
@@ -18,6 +18,7 @@ import propTypes from 'prop-types';
 import { useGetPathwaysQuery } from '../../Services/Pathways';
 import { useIntl } from 'react-intl';
 import { useSelector } from 'react-redux';
+import { workloadQueryBuilder } from '../Common/Tables';
 
 const PathwaysPanel = () => {
   const intl = useIntl();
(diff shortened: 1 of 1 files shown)


src/PresentationalComponents/PathwaysPanel/PathwaysPanel.js
```diff
@@ -18,17 +18,31 @@ import propTypes from 'prop-types';
...
+  const workloads = useSelector(({ filters }) => filters.workloads);
+  const SID = useSelector(({ filters }) => filters.SID);
+  let options = {};
+  selectedTags?.length &&
+    (options = {
+      ...options,
+      ...{ tags: selectedTags.join(',') },
+    });
+  workloads &&
+    (options = { ...options, ...workloadQueryBuilder(workloads, SID) });
```
bastilian, line 39: This could also be written as 

```js
  const options = {
    ...selectedTags?.length > 0 ? { tags: selectedTags.join(',') } : {},
    ...workloads ? workloadQueryBuilder(workloads, SID) : {}
  }
```

or?

adonispuente: ahh I like that better, sending it

bastilian (approved):

PR #1044, opened 2022-11-08
feat(CodeCov): add codecoverage to the repo

Fewwy:
Adding codecoverage to the repo


.babelrc
```diff
@@ -7,8 +7,6 @@
   "plugins": [
     "@babel/plugin-transform-runtime",
     "@babel/plugin-syntax-dynamic-import",
-    "@babel/plugin-proposal-object-rest-spread",
```
Fewwy, line 10: These 2 babel plugins already provided by the babel/preset-env

.github/workflows/cypress.yml
```diff
@@ -9,4 +9,6 @@ jobs:
       - name: Install dependencies
         run: npm install
       - name: Component tests
-        run: npm run test:ct
\ No newline at end of file
+        run: npm run test:ct
+      - name: Create coverage
+        run: npm run coverage
```
gkarat, line 14: I think you'd need to run both jest and cypress tests within the same job. This way, the coverage command will collect coverage results from both tests

coverage.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh -ex
+
+# Merge coverage reports from cypress and jest tests
+mkdir -p reports
+cp cypress-coverage/coverage-final.json reports/from-cypress.json
```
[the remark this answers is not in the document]
Fewwy: It's not included because it's absent. running coverage.sh will result in an error
cp: cannot stat 'jest-coverage/coverage-final.json': No such file or directory

.github/workflows/cypress.yml
gkarat: Yes, looks like that. The same way was done for ocp-advisor.

.github/workflows/cypress.yml
```diff
@@ -9,4 +9,6 @@ jobs:
       - name: Install dependencies
         run: npm install
       - name: Component tests
-        run: npm run test:ct
\ No newline at end of file
+        run: npm run test:ct
```
bastilian, line 12: Two things I'm wondering here. 

1.) This only seems to run the Cypress tests, but do we not want to run bot `jest` and `cypress` test suites, to later combine the reports?

2.) The `nyc` docs suggest that it should be trivial to set up and only require prepending what ever test runner you have with `nyc ...`. Maybe we do not need the babel stuff at all and can try and follow that?[0] and have here as a `run` just `nyc --reporter=lcov npm run test`.


[0] https://github.com/istanbuljs/nyc/blob/master/docs/setup-codecov.md

package.json
```diff
@@ -110,7 +111,7 @@
     "@redhat-cloud-services/frontend-components-config": "^4.6.13",
     "axios": "0.21.4",
     "babel-eslint": "10.1.0",
-    "babel-jest": "27.1.1",
+    "babel-jest": "^23.4.2",
```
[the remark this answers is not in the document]
Fewwy: Mistake, resolved that

.github/workflows/cypress.yml
Fewwy: I'm not sure we should experiment with the NYC in this PR the goal was to get the data about test coverage.
It is made based on the example from vulnerability and ocp advisor, so we're already using the same solution in 2 more repos

Fewwy: @bastilian I'm not sure what we are waiting for. The PR is working, codecov.io have the correct report added. Merge?

bastilian: @Fewwy I have no objections to merge it, it was just a comment/question. But I might not be the best to give an approval, since I only skimmed not reviewed the PR. ;)

Fewwy: @bastilian It's all good! Thanks for the information, that made me think through a couple of things again :)
@gkarat Need your approval

.github/workflows/cypress.yml
```diff
@@ -9,4 +9,8 @@ jobs:
...
+      - name: Jest tests
+        run: npm run test
```
gkarat: Let's make it the same way as was done for vuln4shift-frontend: swap the order of these two steps: so that Jest tests run before component tests.

cypress.config.js
```diff
@@ -1,19 +1,27 @@
...
+const codeCoverageTask = require('@cypress/code-coverage/task');
 
 module.exports = defineConfig({
   viewportWidth: 1000,
   viewportHeight: 660,
   video: false,
   component: {
+    env: {
+      BABEL_ENV: 'component',
+    },
```
gkarat, line 13: Do we need it here since you already provide this env variable at package.json script?

gkarat (commented):
LGTM! Just two minor things to improve and we are good to go

cypress.config.js
Fewwy: Removed it

gkarat (approved):

PR #1045, opened 2022-11-08
fix: Add federated module wrapper for SystemAdvisor

gkarat:
Fixes an issue introduced by https://github.com/RedHatInsights/insights-advisor-frontend/pull/1035.

## Description

This extracts `SystemDetail` federated module to an extra wrapper component. The previous implementation was leading to an error since the `useSelector` hook was trying to access the store before the redux provider was set up. We still require `BaseSystemAdvisor` not to request `entityDetails` via redux store for easier component testing.

## How to test

While deployed locally together with Inventory app (e.g., with test-inv.sh), check:
1) navigate to Advisor system details page and verify the rules table is rendered,
2) navigate to Inventory system details page, Advisor tab, and verify the rules table is rendered.

### DIFF
--- a/config/dev.webpack.config.js
+++ b/config/dev.webpack.config.js
@@ -65,10 +65,7 @@ plugins.push(
       root: resolve(__dirname, '../'),
       exposes: {
         './RootApp': resolve(__dirname, '../src/AppEntry'),
-        './SystemDetail': resolve(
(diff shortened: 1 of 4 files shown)


Fewwy (approved):
tested the branch, infinite load is fixed

PR #1048, opened 2022-11-16
fix(ESSNTL-3781): Fixes filter issues in options object

adonispuente:
This PR introduces the refactor of the options object in the Refactor PR (much more readable), and sends parameter correctly to the api endpoints. 

For testing navigate to to /recommendations, and select an recommendation with a high number of systems.
Here add filters like the os system, name, etc. Also make sure sorting is working and updating properly.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -95,8 +95,7 @@ const Inventory = ({
     selectedIds,
     setFullFilters,
     fullFilters,
-    rule,
-    setFilters
+    rule
   );
 
   const grabPageIds = () => {
--- a/src/PresentationalComponents/Inventory/helpers.js
+++ b/src/PresentationalComponents/Inventory/helpers.js
@@ -68,9 +68,15 @@ export const getEntities =
       SID,
       selectedTags,
     } = config;
-    const sort = `${orderDirection === 'ASC' ? '' : '-'}${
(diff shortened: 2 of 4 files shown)


mkholjuraev (commented):
LGTM! Selection, sorting, and filters (name, tags, OS) are working as expected, and filter chips are displayed correctly. I have noticed that when the table is empty, sorting is still enabled. This leads to a broken page when I sort by **Last seen**. This is a different issue, do you want to work on another PR?

adonispuente:
@mkholjuraev I'd like to keep the options object. I Think it helps make things a little more readable, especially when you have to run it inside multiple functions, it goes inside the handleRefresh and is spread out in the paginatedRequest function.


I've included a commit to fix the issue with the last_seen column breaking things. The value currently isnt accepted by the api request, so now theres a check in place for it. Similar to the one found in SystemsTable

mkholjuraev:
I confirm that the sorting is working when the table is empty. However, I think we can disable sorting completely when the table is empty. There is no usage.

adonispuente:
> I confirm that the sorting is working when the table is empty. However, I think we can disable sorting completely when the table is empty. There is no usage.
@mkholjuraev 
It seems with the current set up that this will require more of a refactor. I'd like to include it in another PR / make a ticket for it

mkholjuraev (approved):
LGTM! @adonispuente thanks!

PR #1052, opened 2022-11-21
feat(GHactions): migrating deploy to the gh actions

Fewwy:
Trying to migrate the deploy settings to the gh actions so they depend on the passing tests
Docs I read
https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-travis-ci-to-github-actions
https://developer.okta.com/blog/2020/05/18/travis-ci-to-github-actions


Fewwy:
> 

The name could be easily updated :) it's not an issue
This PR move deploy to the GH actions, so nobody accidentally deployed something that hadn't passed the tests.
Current solution is less safe :D

.github/workflows/cypress.yml
```diff
@@ -7,6 +7,10 @@ on:
 jobs:
```
gkarat, line 7: Can we also move the Lint step here? Shouldn't we give up the `.travis.yml` completely then? With this PR there is no reason to keep it. 

Furthermore, you have at least remove the deploy step: https://github.com/Fewwy/insights-advisor-frontend/blob/ghactions/.travis.yml#L15. Otherwise, you have a double call of the deploy script.

gkarat (changes requested):

.github/workflows/cypress.yml
Fewwy: Removed the travis.yml and renamed the cypress.yml to deploy.yml + updated it

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
+name: Deploy
```
gkarat, line 1: Can we rename it to something more meaningful? This workflow is not only about deploying. Perhaps something like `pr-ci`. Also could you update the file name accordingly?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm install
+      - name: Jest tests
+        run: npm run test
+      - name: Component tests
+        run: npm run test:ct
+      - name: Create coverage
+        run: npm run coverage
+      - name: Lint
```
gkarat, line 27: We should run linter in the very beginning: for example, after "Install dependencies"

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: npm run lint
+      - name: Deploy
+        run: npm run travis:verify && curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
```
gkarat, line 30: By running `npm run travis:verify` you are basically triggering `npm-run-all travis:build lint test` - almost the same things you do before in this workflow. I would like to remove this call and instead just make a build before requesting the frontend builder:
```suggestion
        run: npm run travis:build && curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
```

gkarat (changes requested):

gkarat (approved):
looks nice 🥇

PR #1054, opened 2022-11-23
fix(ESSNTL-3781-2): Name Param changes if on systems page

adonispuente:
Currently in production when sorting in the systems page, it will send the name param and not filter down the table. This PR changes that so that the correct param is sent for the correct table.

In the systems page it needs to send display_name.
In recommendations table in needs to send text.
in pathways table it also needs to send name

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -285,7 +285,8 @@ const SystemsTable = () => {
             filters,
             selectedTags,
             workloads,
-            SID
+            SID,
+            true
           );
           const fetchedSystems = (await Get(SYSTEMS_FETCH_URL, {}, options))
             ?.data;
--- a/src/PresentationalComponents/helper.js
+++ b/src/PresentationalComponents/helper.js
@@ -9,7 +9,8 @@ export const createOptions = (
   filters,
   selectedTags,
   workloads,
-  SID
+  SID,
+  systemsPage
 ) => {
   let options = {
     ...advisorFilters,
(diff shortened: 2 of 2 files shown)


mkholjuraev:
The recommendation systems table is not getting affected by the Name filter. It sends an empty name to API.

adonispuente:
@mkholjuraev I believe, the recommendations table is the RulesTable.js. It should be sending a text param. That table does not use the createOptions function and was not touched in the refactor. 

It should still be filtering down fine. Is it broken?

mkholjuraev:
@adonispuente the issue is on recommendation detail page, systems tab. Sorry for confusing you by saying recommendation systems.

mkholjuraev (approved):
LGTM! Observed the following params on those pages:
1. Systems table -  display_name.
2. Affected systems  - name
3. Pathway system  -display_name (You wrote it should pass name instead of display_name in the PR comment. But with this PR, it is working as expected)

All those tables are working as expected. However, I believe that the difference in params for **Name** filter will continue causing issues unless we get rid of the difference. We need to talk to the engine team and ask them to have either `display_name` for all page filters or `name`.


There is one more thing that needs to be fixed in the upcoming commits:
image

src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -285,7 +285,8 @@ const SystemsTable = () => {
             filters,
             selectedTags,
             workloads,
-            SID
+            SID,
+            true
```
mkholjuraev, line 289: Can we refactor this function so that it does not use a constant boolean for detecting the systems page? I am having difficulties on how to write it down here, feel free to ping me if more explanation is needed.

mkholjuraev:
I failed to notice that the name of the filter chip is not the same as the filter. The filter name is 'Name' while the chip's name is 'Display name'

gkarat (approved):
Compared requests made running this PR and on the stage deployment: the PR fixes the bug and now the HTTP request is https://stage.foo.redhat.com:1337/api/insights/v1/system/?sort=-last_seen&limit=20&offset=0&hits=all&display_name=foobar, the broken one was https://console.stage.redhat.com/api/insights/v1/system/?sort=-last_seen&limit=20&offset=0&hits=all&name=foobar, lgtm ✅ 

I think we need to prioritize our work on the tests coverage and implement some verification functions for API-paginated tables like this.

PR #1059, opened 2022-11-29
feat(RecsListTests): sorting tests

Fewwy:
Need advice, during the development, I discovered that if we double-click on the sorting button, URL addresses do not contain minus sign before the sorting parameter. When I added the console.log(window.location.href); to the "descending" part of the IF statement, it showed the same URL address as in ascending.

If we can solve this issue - we can add these tests to the other tables where we utilize sorting with API requests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -156,7 +156,8 @@
     ],
     "exclude": [
       "src/**/*.spec.ct.js",
-      "src/**/*.test.js"
+      "src/**/*.test.js",
+      "src/**/*.cy.js"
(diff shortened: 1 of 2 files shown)


Fewwy:
Hmm, I checked the PR that @gkarat merged yesterday (he added tests to the same file), and test coverage increased. Now Codecov tells me that adding tests reduced the coverage

adonispuente (approved):
PR works as intended, LGTM!

PR #1061, opened 2022-11-29
fix(ADVISOR-2894): updates sorting conditions in SystemAdvisor page.

adonispuente:
# Description
Sorting results are altered when going from recommendations > rec details > affected systems.
The first value is never moved regardless of what you sort by

Associated Jira ticket: # (ADVISOR-2894)
This PR updates the conditions the table needs to sort the rule as the first item on the table.

# How to test the PR
There are 2 scenarios to test
Advisor recommendations -> rec details -> Affected system: first in the table should be the recommendation in question, expanded. Then you can sort normally
&
Advisor systems -> system details: all recommendations are collapsed. You can sort normally

# Before the change
The first item would never move when coming to the page from that path

# After the change
The recommendation is at the top on page load, but you can sort the page normally afterwards

# Dependent work link

# Checklist:


gkarat (approved):
Run locally, the sorting works great 🎉  LGTM!

PR #1068, opened 2022-12-06
fix(ADVISOR-2915): Updates filters check for RulesTable

adonispuente:
# Description

Associated Jira ticket: # ([ADVISOR-2915](https://issues.redhat.com/browse/ADVISOR-2915))
It seems Advisor rules table was originally built to update filters based off urls on page load. Different tables take in differing parameter values in advisor (this is a known issue).

# How to test the PR

Run advisor against prod env via :
CHROME_ENV=prod-stable npm run start:proxy
Access https://console.redhat.com/insights/dashboard
In the menu select Advisor > Recommendations
In the global filter select "Microsoft SQL"
Select a recommendation to show the affected systems
Click on the "Go back" button in the browser

# Before the change
The table would break, in the network tab it would yell about an incorrect sorting param passed in

# After the change
It passes, making the default '-total_risk'

Notes: In the refactor PR I will try and optimize this page further.

# Checklist:

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.cy.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.cy.js
@@ -392,3 +392,42 @@ describe('Sorting', () => {
     });
   });
 });
+
+const urlParamsList = [
+  'impacting=true&rule_status=enabled&sort=-total_risk&limit=20&offset=0#SIDs=&tags=',
+  'impacting=false&rule_status=enabled&sort=-recommendation_level&limit=20&offset=0#SIDs=&tags=',
+];
+
+urlParamsList.forEach((urlParams, index) => {
+  describe(`pre-filled url search parameters ${index}`, () => {
+    beforeEach(() => {
+      cy.intercept('*', {
+        statusCode: 201,
+        body: {
+          ...fixtures,
+        },
+      }).as('call');
+
+      cy.mount(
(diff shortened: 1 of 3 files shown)


mkholjuraev:
This works as expected, but what if some other sorting param that is not supported in the API is applied? 

How about utilizing existing [sort indices](https://github.com/RedHatInsights/insights-advisor-frontend/blob/75e0fa8302f53b7ec9b6077ba35c0cde45a542da/src/PresentationalComponents/RulesTable/RulesTable.js#L151) to make sure a sort param that exists in that map is applied?

adonispuente:
@mkholjuraev I might be able to make something happen with jest test, but for the sort indices , on page load there are filters being built via the url, not from anything on the table. I believe the thinking was based of the pathways table, but when navigating to /recommendations/ the url sort param is meant for the pathways table and not the rulesTable. Im not sure how I can use the sort Indices to correct this problem. 

An idea I had for the refactor is to make both of these tables have different routes

adonispuente:
@mkholjuraev the changes you requested are up ^

Fewwy (approved):
Tested it locally, works fine. LGTM

mkholjuraev (approved):
LGTM! sorry for late reply.

adonispuente:
@mkholjuraev @Fewwy Thank you both for reviewing !

PR #1072, opened 2022-12-13
fix(RHCLOUD-23202): Do not re-create sotre on each render

karelhala:
# Description

Associated Jira ticket: #RHCLOUD-23202

There's an error with global filter tags not being applied when selecting. This is caused by re-creating store on each AppEntry render. The application can render multiple times (PF contexts, navigation, etc.) and that can invalidate data stored in the redux state. This PR fixes it by using `useMemo` to create the store only on initial render.

# How to test the PR

Run Advisor app locally and select tag from global filter.


adonispuente (approved):
PR works as intended, LGTM! Thank you!

PR #1074, opened 2022-12-22
fix(SystemAdvisor): Request all kcs articles

gkarat:
# Description

Associated Jira ticket: https://issues.redhat.com/browse/ADVISOR-2948.

SystemAdvisor table renders so-called "Knowledgebase article" section where it also inserts a link to access.redhat.com with info relevant for particular rule. To learn the appropriate URL and publish title, it makes additional request to access.redhat.com API and provides all necessary rule (node) IDs. The issue is that we were providing this request also the parameter `rows` and it was set to the number of rule IDs. However, the API can return more results, and in some edge cases (see Jira), there can be some rules left out from the response, and the table was breaking. By removing `rows` parameter, we are making sure we receive kcs links for all node IDs.

# How to test the PR

The scenario is hard to simulate, but I checked the customer case and simulated the API fixtures that lead to this issue (check `fixturesKcs`) and wrote some cypress tests that have to pass. 

# Before the change

The table would break if it hasn't fetched all URLs and titles for all node ids.

# After the change

The response from API should contain all URLs for all node ids.

# Checklist:

- [x] Tests for the changes have been added

### DIFF
--- a/cypress/fixtures/kcs.json
+++ b/cypress/fixtures/kcs.json
@@ -1,0 +1,59 @@
+[
+  {
+    "publishedTitle": "Scelerisque odio diam placerat semper pretium etiam, sollicitudin nisl enim risus lacinia.",
+    "id": "300533",
+    "view_uri": "http://foo.bar/articles/300533"
+  },
+  {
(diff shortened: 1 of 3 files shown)


mkholjuraev (approved):
LGTM!

PR #1083, opened 2023-01-17
fix(codecov): added code coverage config

Fewwy:
# Description
I encountered the issue of "codecov/patch" breaking while I was working on another PR ( see the screenshot)
To fix that issue and incorrect behavior I added a code coverage config

image

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,16 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: 0%
+        threshold: 10%
+        patch: false
+        flags:
+          - unit
+       # advanced settings
+        branches:
+          - master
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -0,0 +1,16 @@
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: 0%
```
mkholjuraev, line 6: Maybe it is good to have a target?

mkholjuraev (approved):
LGTM!

.codecov.yml
Fewwy: @mkholjuraev as far as I know this won't block PR's, and the goal was that :) 
we can control the % ourselves without any blocking codecov/patch things

PR #1094, opened 2023-01-31
fix(ADVISOR-2965): adds check against url

adonispuente:
Bandaid fix for advisors sorting issue on page load. Sometimes the value is an array and others a string. This adds conditions to that check. 
Still investigating the real issue at hand but I wanted to send this up as its listed other major incidents. 

# How to test the PR

go to Recommendations page
apply a filter (NOT name filter)
copy the URL
navigate to the URL in a new tab

# Before the change
breaks on sort in url

# After the change
Does not break

# Dependent work link

# Checklist:


Fewwy:
@adonispuente Tested it against prod-beta env, LGTM

Fewwy (approved):

PR #1104, opened 2023-04-06
fix(ADVISOR-3032)Update empty state when rule is not found

adonispuente:
# Description
Updates empty state behavior so that when a user goes to a bookmark of a rule that no longer exist, they are greeted with the correct page. 

Associated Jira ticket: # (ADVISOR-3032)
# How to test the PR
Head to https://stage.foo.redhat.com:1337/beta/insights/advisor/recommendations/whatever#SIDs=&tags= 
Observe that the page does not error out but instead shows this image
Screen Shot 2023-04-06 at 5 04 49 PM
# Before the change
The page would crash and get  "Something went wrong". 

# After the change
Page does not crash anymore.

# Dependent work link

# Checklist:

### DIFF
--- a/src/SmartComponents/Recs/Details.js
+++ b/src/SmartComponents/Recs/Details.js
@@ -107,41 +107,40 @@ const OverviewDetails = () => {
 
   return (
     <React.Fragment>
-      {viewSystemsModalOpen && !isError && (
-        <ViewHostAcks
-          handleModalToggle={(toggleModal) =>
-            setViewSystemsModalOpen(toggleModal)
-          }
-          isModalOpen={viewSystemsModalOpen}
-          afterFn={() => refetch()}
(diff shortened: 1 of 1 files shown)


adonispuente:
ci fails because the container is already in use

leSamo (approved):
Works well, but the boolean conditions are becoming a bit of a spaghetti :( can they maybe be reorganized a bit?

adonispuente:
@leSamo I've cleaned it up a bit, it can definitely use some more love. The entire page is a huge amount of conditionals. Ideally I'd like to do a refactor of the page as a whole. I had done some in the really large refactor PR but it was too big for anytime to have time to completely look at. Definitely going to work on sending up small bits of refactored pieces moving forward

PR #1125, opened 2023-05-23
chore(Refactor): Reorganize RulesTable component

adonispuente:
Learning from my last large attempt at refactoring, This PR reorganizes just the RulesTable component. There should be no changes in functionality. This simply cleans up the RulesTable component so that its much easier to navigate. The file was 800+ lines of code and it caused of lot of jumping around when trying to debug the table. This PR cuts it in half and utilizes the helpers.js file

# Checklist:

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -1,18 +1,13 @@
 import './_RulesTable.scss';
-
 import * as AppConstants from '../../AppConstants';
-
 import { DEBOUNCE_DELAY, FILTER_CATEGORIES as FC } from '../../AppConstants';
-import { Link, useLocation } from 'react-router-dom';
+import { useLocation } from 'react-router-dom';
 import {
   Pagination,
   PaginationVariant,
 } from '@patternfly/react-core/dist/esm/components/Pagination/Pagination';
 import React, { useEffect, useState } from 'react';
-import {
-  Stack,
-  StackItem,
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -173,60 +153,12 @@ const RulesTable = ({ isTabActive }) => {
     setFilters({ ...filters, offset: newOffset });
   };
 
-  const toggleRulesDisabled = (rule_status) => {
```
mkholjuraev, line 176: I would vote for leaving functions, like this, that are specific to a component inside. This function deals with updating a state only. Functions that can be reused in other components are great to be refactored

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -284,161 +234,13 @@ const RulesTable = ({ isTabActive }) => {
       if (rules.data.length === 0) {
         setRows(emptyRows(filters, toggleRulesDisabled));
       } else {
-        const rows = rules.data.flatMap((value, key) => [
```
mkholjuraev, line 287: Happy to see this got refactored!

mkholjuraev (commented):
Great for initial refactoring. I believe that in the next phase, we can try to see how those extracted functions could be minimized.

src/PresentationalComponents/RulesTable/RulesTable.js
adonispuente: This makes sense, I've moved it back and just passed it into the filter config for the onChange function

PR #1127, opened 2023-05-26
chore(deps): cypress update to v 12.3

Fewwy:
Had to update cypress to the newer version because we have to add overrideChrome and it breaks cypress version <11


AsToNlele (approved):
LGTM... again :smile:

PR #1130, opened 2023-06-01
feat(ESSNTL-3715): added group columns and a check for group columns

Fewwy:
# Description
First part of the 3715 jira ticket

https://issues.redhat.com/browse/ESSNTL-3715

# How to test the PR
Check if the Group column is present in the systems table
Turn off the https://insights-stage.unleash.devshift.net/projects/default/features/hbi.ui.inventory-groups
restart a page 1 or 2 times to make sure that the feature flag is turned off
Group column should be missing

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTableAssets.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTableAssets.js
@@ -22,6 +22,7 @@ export const systemsTableColumns = (intl) => [
   },
   {
     key: 'groups',
+    requiresDefault: true,
   },
   {
     key: 'tags',
--- a/src/PresentationalComponents/SystemsTable/SystemsTableAssets.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTableAssets.js
@@ -21,6 +21,9 @@ export const systemsTableColumns = (intl) => [
     ),
   },
   {
+    key: 'groups',
+  },
+  {
     key: 'tags',
   },
   {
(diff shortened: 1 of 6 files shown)


src/PresentationalComponents/SystemsTable/createColumns.js
gkarat, line 1: As for our previous discussion, this solution makes sense. But I propose slightly different one that could make the logic more agnostic to the existence of the groups feature:

Can we instead add one extra flag to `SystemTableAssets` list like `requiresDefault`? If it's set to true, then the function will filter out this column if it can't find it available in `defaultColumns`. If set to false or not set explicitly, keep it. See my second comment on how I'd rewrite `SystemTableAssets` in that case. And here is the code of function I am thinking of:

```
export const createColumns = (defaultColumns, columns) =>
  columns
    .map((column) => {
      const correspondingColumn = defaultColumns.find(
        (defaultColumn) => defaultColumn.key === column.key
      );

      return column.requiresDefault && correspondingColumn === undefined
        ? undefined
        : { ...column, ...correspondingColumn };
    })
    .filter(Boolean);
```

src/PresentationalComponents/SystemsTable/SystemsTableAssets.js
```diff
@@ -20,6 +20,9 @@ export const systemsTableColumns = (intl) => [
       </React.Fragment>
     ),
   },
+  {
+    key: 'groups',
+  },
```
gkarat, line 25: see my previous comment

```suggestion
  {
    key: 'groups',
    requiresDefault: true
  },
```

Fewwy: Added it :)

src/PresentationalComponents/SystemsTable/createColumns.js
Fewwy: Added it :)

PR #1132, opened 2023-06-06
feat(ESSNTL-3715): Support filtering by group, add the column

Fewwy:
Implements https://issues.redhat.com/browse/ESSNTL-3715.

This makes sure the Advisor /systems page supports filtering by groups and that the appropriate column is showing a group for each host.

## How to test

1. Navigate to /systems.
2. Check that the Groups filter is available in the filters toolbar, and the groups name are equal to the ones you can find on the Inventory systems page.
3. Select any filter(s), and then, in the Network tab, check that the request to /api/insights/v1/system has the group names as query parameters. **You shouldn't see data changed in the table because the groups parameter is not yet supported by back end.** 
4. Find any host which is in a group and try filter it in the table. Check that the Group column is rendered correctly and you can click on the group name and be redirected to /inventory/groups/%id.
5. Make sure filter chips are rendered and the table's dynamic URL is updated correctly.

## Screenshots

image


Fewwy:
@gkarat LGTM

PR #1146, opened 2023-06-28
chore(deps): remove codecov patch

Fewwy:
turn off codecov patch because its keep popping up and throwing CI fails in our PR's


mkholjuraev (approved):
LGTM!

PR #1157, opened 2023-07-28
Reverts all the pdf service migration work

mkholjuraev:
This reverts all the work related to the new PDF service migration until the service works properly. When merging, we need to make sure to squash into one commit, so that we can easily bring the work back.


adonispuente (approved):

adonispuente:
Right now, since master branch still has the commits, we WOULD be able to bring the commits back easily. It would only be an issue if we started off of another branch, but since we push from master, this should not be a problem. However if its your preference to do this that should be fine

PR #1159, opened 2023-07-31
fix(ADVISOR-3139): Resolve system pdf export error

adonispuente:
# Description

Associated Jira ticket: # ([issue](https://issues.redhat.com/browse/ADVISOR-3139))
This is a temporary fix while we wait for the real pdf migration 

# How to test the PR

Navigate to Systems page and clicked the export button.
Once the PDF option loads, select it and verify that there is no 500 error. The backend doesnt support the tags field. Normally a refactor for this would be passed through, but since theres a PR for migrating to the new pdf service I'd like to pass this in until that gets QA'd and eventually removes this commit all together. 
# Before the change
Systems pdf export breaks

# After the change
Doesnt break

# Dependent work link

# Checklist:

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -5,7 +5,7 @@ import {
   SYSTEM_FILTER_CATEGORIES as SFC,
   SYSTEMS_FETCH_URL,
 } from '../../AppConstants';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { TableVariant } from '@patternfly/react-table';
 import {
   filterFetchBuilder,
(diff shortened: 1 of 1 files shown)


src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -219,7 +219,8 @@ const SystemsTable = () => {
     urlBuilder(combinedFilters, selectedTags);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [selectedTags]);
-
+  let pdfFilters = filterFetchBuilder(filters);
+  delete pdfFilters.tags;
```
bastilian, line 223: It might make sense to put this into a `useMemo`.

adonispuente: Good catch

bastilian: More like:

```js
const pdfFilters = useMemo(() => {
 const { _tags, ...filters } = filterFetchBuilder(filters);
 return filters;
}, [filters]);
```

adonispuente: Im not too familiar with this syntax, but this breaks the code. 
Tags is undefined and that breaks the page

bastilian (approved):
Let's get this merged! Thank you @adonispuente!

PR #1166, opened 2023-08-21
Advisor Systems page and Inventory groups enhancements

gkarat:
Implements https://issues.redhat.com/browse/ESSNTL-3715.

- Make sure that the systems table (/advisor/systems) uses "groups" parameter to filter systems by their groups,
- Change the Group column values to plain text instead of links.

## How to test

Go to /advisor/systems, and check that you are able to filter by groups and also can see to which group each hosts relates to (N/A if not in any group).

### DIFF
--- a/src/PresentationalComponents/SystemsTable/createColumns.js
+++ b/src/PresentationalComponents/SystemsTable/createColumns.js
@@ -1,6 +1,3 @@
-import React from 'react';
-import isEmpty from 'lodash/isEmpty';
-
 export const createColumns = (defaultColumns, columns) =>
   columns
     .map((column) => {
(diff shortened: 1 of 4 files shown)


src/PresentationalComponents/SystemsTable/createColumns.js
```diff
@@ -13,23 +10,6 @@ export const createColumns = (defaultColumns, columns) =>
         : {
             ...column,
             ...correspondingColumn,
-            ...(column.key === 'groups'
-              ? {
-                  renderFunc: (groups) =>
-                    isEmpty(groups) ? (
```
mkholjuraev, line 19: If there is no group, we need to show N/A. With this change, I believe this will not be the case ?

gkarat: @mkholjuraev, it will be already done by Inventory filter config (the default one). The removed code was making sure that we render the group names as links - this is no longer required.

gkarat: You also can run it locally and see that N/A is rendered for empty groups.

mkholjuraev (approved):
LGTM! works as described, @gkarat thank you!

PR #1168, opened 2023-09-08
Fix(RHINENG-1847): expand the correct selected recommendation

mhuth:
Fixes https://issues.redhat.com/browse/RHINENG-1847

Selecting an affected host for a particular recommendation, the wrong recommendation is expanded.  

Screenshot showing the recommendation and the affected hosts ...

Screenshot from 2023-09-08 13-06-35

Before the fix - the incorrect recommendation is expanded for the host:

Screenshot from 2023-09-08 13-03-46

After the fix - the correct recommendation is expanded for the host:

Screenshot from 2023-09-08 13-06-45


AsToNlele (approved):
LGTM, thank you! :smile:

PR #1189, opened 2023-11-02
fix(SystemsTable): RHIN-338 - Remove custom OS filter handling

bastilian:
# Description

Associated Jira ticket: #RHIN-338

This migrates the SystemsTable to use the default OS filter provided by the InventoryTable and let it handle the filter selection.

# Before the change

https://github.com/RedHatInsights/insights-advisor-frontend/assets/7757/e7e2d0a1-49ad-4cf4-a086-b91349e11cff

# After the change

https://github.com/RedHatInsights/insights-advisor-frontend/assets/7757/1400f080-c693-47ca-b3c9-e465490fbb6d

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -20,7 +20,6 @@ import { updateReducers } from '../../Store';
 import { useIntl } from 'react-intl';
 import downloadReport from '../Common/DownloadHelper';
 import useBulkSelect from './Hooks/useBulkSelect';
-import { Spinner } from '@scalprum/react-core';
 
 const Inventory = ({
   tableProps,
@@ -436,7 +435,6 @@ const Inventory = ({
           ],
         }}
         {...toolbarProps}
-        fallback={Spinner}
         onLoad={({
           mergeWithEntities,
           INVENTORY_ACTION_TYPES,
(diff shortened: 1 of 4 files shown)


adonispuente:
Screen Shot 2023-11-02 at 11 57 13 AM

Located via Recommendations > clicking a rec > 
There is an additional place the table is used. In the HybridInventoryTab.js file.

bastilian:
@adonispuente Thank you for reviewing and testing. And catching that! :) The InventoryTable there is now migrated too.

adonispuente (approved):
Everything looks good and works as intended! The SystemAdvisor cypress test just need to be updated

package.json
```diff
@@ -76,7 +76,6 @@
     "@redhat-cloud-services/frontend-components-translations": "3.2.6",
     "@redhat-cloud-services/frontend-components-utilities": "^3.7.3",
     "@reduxjs/toolkit": "1.6.1",
-    "@scalprum/react-core": "^0.1.9",
```
gkarat, line 79: It looks like this change is causing test failure. RemediationButton from FEC imports @scalprum/react-core, however, the package doesn't list scalprum as a dependency in its package.json. I think we need to keep this dependency in advisor for now, but certainly this is the bug in RemediationButton component. I will open an issue.

mkholjuraev (approved):
LGTM! tested locally, did not face issues

PR #1194, opened 2023-11-14
fix(ImmutableDevices): pass edge filer into custom filter prop

mkholjuraev:
# Description

Associated Jira ticket: # NO TICKET

This fixes the broken immutable devices tab on the Recommendation details page. The issue happens when you filter an already empty table and then click on the 'Reset filter' button. This was due to the fact that the `'filter[system_profile][host_type]': 'edge'` filter was placed inside a state. When the Reset button was clicked, that filter was gone and the API request brought incorrect data. Now, it is outside of useState, and thus API requests have this filter in place always.

# How to test the PR

1. Run the PR
2. Navigate to the Recommendation detail page that has no immutable devices (empty table)
3. Filter the table with Name fitler
4. Click on reset button
5. Observe that the table is not broken. (Make sure to repeat this multiple times)

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/SmartComponents/HybridInventoryTabs/ImmutableDevices.js
+++ b/src/SmartComponents/HybridInventoryTabs/ImmutableDevices.js
@@ -27,9 +27,6 @@ const ImmutableDevices = ({ rule, pathway, selectedTags }) => {
     offset: 0,
     sort: '-last_seen',
     name: '',
-    ...(isEdgeParityEnabled
-      ? { 'filter[system_profile][host_type]': 'edge' }
-      : {}),
   });
 
   const handleRefresh = (options) => {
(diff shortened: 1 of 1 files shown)


src/SmartComponents/HybridInventoryTabs/ImmutableDevices.js
```diff
@@ -135,7 +132,12 @@ const ImmutableDevices = ({ rule, pathway, selectedTags }) => {
       }}
       key="inventory"
       customFilters={{
-        advisorFilters: filters,
+        advisorFilters: {
+          ...filters,
+          ...(isEdgeParityEnabled
+            ? { 'filter[system_profile][host_type]': 'edge' }
+            : {}),
+        },
```
gkarat, line 140: nit-pick: do you mind extracting this to a function? and we could call it like "buildAdvisorEdgeFilters" to better illustrate that this parameter is (and has to be) wired into the filters.

gkarat (approved):
LGTM! Thanks @mkholjuraev. Though I couldn't reproduce this bug for some reason in stage.

PR #1196, opened 2023-11-15
feat(RHINENG-3064): Add Modified Table title of Rec Details Title

adonispuente:
This PR institues https://issues.redhat.com/browse/RHINENG-3064 and adds a different title to Rec details page. 
Acceptance criteria:

When there is an immutable system, the title above the table should include total number
Total number should count both conventional and immutable systems
when there is no immutable system, the title should have no change and the total number

### DIFF
--- a/src/SmartComponents/Recs/Details.js
+++ b/src/SmartComponents/Recs/Details.js
@@ -2,7 +2,7 @@ import './Details.scss';
 
 import { PERMS, UI_BASE } from '../../AppConstants';
 import messages from '../../Messages';
-import React, { useEffect, useState } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import propTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
 import { useIntl } from 'react-intl';
@@ -28,10 +28,11 @@ import { cveToRuleid } from '../../cveToRuleid.js';
 import { useGetRecAcksQuery } from '../../Services/Acks';
 import { useGetRecQuery } from '../../Services/Recs';
 import { useGetTopicsQuery } from '../../Services/Topics';
-import { enableRule, bulkHostActions } from './helpers';
+import { enableRule, bulkHostActions, edgeSystemsCheck } from './helpers';
 import { DetailsRules } from './DetailsRules';
 import { useChrome } from '@redhat-cloud-services/frontend-components/useChrome';
 import HybridInventory from '../HybridInventoryTabs/HybridInventoryTabs';
+import { AccountStatContext } from '../../ZeroStateWrapper.js';
 
 const OverviewDetails = ({ isImmutableTabOpen }) => {
   const intl = useIntl();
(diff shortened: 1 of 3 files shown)


src/SmartComponents/Recs/helpers.js
```diff
@@ -72,3 +72,27 @@ export const bulkHostActions = async ({
     });
   }
 };
+
+export const edgeSystemsCheck = async (ruleId, setSystemsCount) => {
```
mkholjuraev, line 76: How about converting this function to return the number of conventional and edge systems count? With this would be able use those numbers in HybridInventoryTab component also to switch to edge tab when there is no conventional system, but there is edge automatically.

src/SmartComponents/Recs/Details.js
```diff
@@ -92,6 +95,10 @@ const OverviewDetails = ({ isImmutableTabOpen }) => {
     }
   }, [chrome, intl, rule.description, ruleId]);
 
+  useEffect(() => {
+    edgeSystemsCheck(ruleId, setSystemsCount);
```
mkholjuraev, line 99: Let's put this work behind feature flag `advisor.edge_parity`

mkholjuraev (commented):
Works as expected

src/SmartComponents/Recs/helpers.js
adonispuente: I can add it in, but currently it doesnt accept that prop. We'd have to make an PR to update HybridInventory to accept it

mkholjuraev: [hasConventionalSystems](https://github.com/RedHatInsights/insights-inventory-frontend/blob/e3e7158e47efbad4a3ab2df9b7d133c2fcd01e8e/src/components/InventoryTabs/HybridInventoryTabs.js#L23) props are designed for this. But yeah now I think that prop might have been more meaningful. It is now a bit confusing.

src/SmartComponents/Recs/helpers.js
```diff
@@ -72,3 +72,34 @@ export const bulkHostActions = async ({
     });
   }
 };
+
+export const edgeSystemsCheck = async (
+  ruleId,
```
bastilian, line 77: ^ +1 We should ad a test for this helper.

src/SmartComponents/Recs/helpers.js
```diff
@@ -72,3 +72,34 @@ export const bulkHostActions = async ({
...
+      .get(
+        `/api/insights/v1/rule/${ruleId}/systems_detail/?filter[system_profile][host_type][nil]=true`
+      )
+      .then(({ data }) => {
+        count = count += data.meta.count;
+        setConventionalSystemsCount(data.meta.count);
+      });
+    await axios
+      .get(
+        `/api/insights/v1/rule/${ruleId}/systems_detail/?filter[system_profile][host_type][edge]=true`
```
mkholjuraev, line 94: @adonispuente I missed this: the filter syntax should be 
```suggestion
        `/api/insights/v1/rule/${ruleId}/systems_detail/?filter[system_profile][host_type]=edge`
```

Currently, the API request always returns 0 immutable count. Also, it looks like there should be a loading state as the total number stays in 0 for a while and then gets updated after API requests are finished

mkholjuraev (changes requested):

src/SmartComponents/Recs/helpers.test.js
```diff
@@ -0,0 +1,27 @@
+import * as axios from 'axios';
+// import MockAdapter from 'axios-mock-adapter';
```
mkholjuraev, line 2: We can safely remove this :)

mkholjuraev (approved):
LGTM! Thank you @adonispuente. tested locally, seems everything is fine now.

mkholjuraev (approved):

PR #1201, opened 2023-11-22
fix(ImmutableDevices): column spacing is fixed by extending column props

mkholjuraev:
# Description

Fixes spacing between groups and tags columns by extending the column prop with default prop in mergeAppColumns function.

# How to test the PR

# Before the change

image

# After the change

image

# Dependent work link

# Checklist:

- [x] Tests for the changes have been added

### DIFF
--- a/src/SmartComponents/HybridInventoryTabs/helpers.test.js
+++ b/src/SmartComponents/HybridInventoryTabs/helpers.test.js
@@ -1,16 +1,32 @@
 import { renderHook } from '@testing-library/react-hooks';
 import { act } from '@testing-library/react';
-import { useGetEntities, useActionResolver, mergeAppColumns } from './helpers';
+import {
+  useGetEntities,
+  useActionResolver,
+  mergeAppColumns,
+  useOnLoad,
+} from './helpers';
 import { Get, Post } from '../../Utilities/Api';
 import inventoryData from './fixtures/inventoryData.json';
 import advisorPathwayData from './fixtures/advisorPathwayData.json';
 import advisorRecommendationData from './fixtures/advisorRecommendationData.json';
 import edgeData from './fixtures/edgeData.json';
+import { updateReducers } from '../../Store';
 
 jest.mock('../../Utilities/Api', () => ({
   ...jest.requireActual('../../Utilities/Api'),
   Get: jest.fn(() => Promise.resolve({})),
   Post: jest.fn(() => Promise.resolve({})),
+}));
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useStore: jest.fn(() => ({ replaceReducer: jest.fn() })),
+}));
+
+jest.mock('../../Store', () => ({
+  ...jest.requireActual('../../Store'),
+  updateReducers: jest.fn(() => {}),
 }));
 
 Post.mockReturnValue(edgeData);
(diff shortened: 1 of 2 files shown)


src/SmartComponents/HybridInventoryTabs/helpers.js
```diff
@@ -87,3 +91,42 @@ export const useActionResolver = (handleModalToggle) =>
     ],
     []
   );
+
+export const useOnLoad = (filters) => {
+  const store = useStore();
+  return useCallback(
+    ({ mergeWithEntities, INVENTORY_ACTION_TYPES, mergeWithDetail }) => {
+      store.replaceReducer(
+        updateReducers({
```
bastilian, line 100: ^ Can this be tested as it is pointed out here?

mkholjuraev: @bastilian thank you for the review. I thought about this, but could not find a meaningful way of covering with tests. Happy to implement some suggestions.

bastilian: @mkholjuraev You could make the AsyncComponent mock in the ImmutableDevices test call the function provided via the `onLoad`. Than this code would be at least called. You could also add a separate test for the hook and assert that it does correctly "merge" and provides proper arguments to the `mergeWithEntities` function.

src/SmartComponents/HybridInventoryTabs/helpers.test.js
```diff
@@ -144,3 +144,34 @@ describe('useActionResolver', () => {
...
+    );
+    expect(system_profile.props).toEqual({ width: 15, isStatic: true });
+  });
+  test('Should extend groups column props to disable sorting', () => {
+    const result = mergeAppColumns(defaultColumns);
+
+    const groups = result.find((column) => column.key === 'groups');
+    expect(groups.props).toEqual({ width: 15, isStatic: true });
+  });
+});
```
gkarat, line 177: Would be great if we could split the hook and tests related work into a separate commit. It confused me a bit why it's here whereas the PR only states that you fix the columns width

src/SmartComponents/HybridInventoryTabs/helpers.js
```diff
@@ -87,3 +91,42 @@ export const useActionResolver = (handleModalToggle) =>
...
+};
+
+export const mergeAppColumns = (defaultColumns) => {
+  const lastSeenColumn = defaultColumns.find(({ key }) => key === 'updated');
+  const impacted_date = {
+    key: 'impacted_date',
+    title: 'First Impacted',
+    sortKey: 'impacted_date',
+    transforms: [sortable, wrappable],
+    props: { width: 15 },
```
gkarat, line 120: I am not sure this is the standard way we use to change the columns width. Shouldn't `cellWidth` transform property serve you better? https://www.patternfly.org/2022.11/components/table/react-legacy#controlling-text

src/SmartComponents/HybridInventoryTabs/helpers.test.js
mkholjuraev: Hmm, I am not sure if this gives us any benefit? The PR is not large and furthermore once approved I will squash the PR into the master.

src/SmartComponents/HybridInventoryTabs/helpers.js
mkholjuraev: As you can see these columns are passed to the inventory table in the middle. InventoTable then it gets mapped to cellWith down the line here: https://github.com/RedHatInsights/insights-inventory-frontend/blob/e9af8e577b5c088d4f377b11218393e7918ca9af/src/components/InventoryTable/helpers.js#L120

mkholjuraev:
> How the Group and Tags columns gap was increased if you didn't modify their props in this PR anyhow? 🤔

They were affected because immutable uses groups column definitions coming as a part of defaultColumns and I was wrongly mutated it completely to make it static here: https://github.com/RedHatInsights/insights-advisor-frontend/pull/1201/files#diff-1c08cf0a43c16938d52ff334bd629f63d3efb4de80c72f9480b903f787893283L111. This PR fixes it by extending incoming column props with `isStatic: true`

src/SmartComponents/HybridInventoryTabs/helpers.test.js
gkarat: OK, it was a nit-pick rather, feel free to move on

gkarat (dismissed):
LGTM. this works as expected. I think it would be nice to improve tests as @bastilian suggested: https://github.com/RedHatInsights/insights-advisor-frontend/pull/1201#discussion_r1407622531. But giving this approval

mkholjuraev:
@gkarat thank you for the review. I will add the suggested tests and then merge this.

LiorKGOW:
Hey @mkholjuraev could you add the Jira ticket number please ?

LiorKGOW:
Is this commit ready for a release @mkholjuraev ?

PR #1202, opened 2023-11-24
feat(RHINENG-5019): Column modifications to RecommendationsSystems

mhuth:
https://issues.redhat.com/browse/RHINENG-5019

Adds the Group column to the RecommendationsSystems table and enables sorting on the OS column.  However sorting on Group column isn't enabled yet because the API doesn't support it yet.

Before PR (no group column and no sorting on OS): 
Screenshot from 2023-11-25 03-04-40

After PR:
Screenshot from 2023-11-25 03-04-34

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -278,6 +278,8 @@ const Inventory = ({
         ({ key }) => key === 'system_profile'
       );
       let tags = defaultColumns.filter(({ key }) => key === 'tags');
+      let groups = defaultColumns.filter(({ key }) => key === 'groups');
+
       //Link to the Systems in the Recommendation details table and Pathway details table
       displayName = {
         ...displayName[0],
@@ -302,20 +304,30 @@ const Inventory = ({
       lastSeenColumn = {
         ...lastSeenColumn[0],
         transforms: [sortable, wrappable],
-        props: { width: 20 },
       };
 
       systemProfile = {
         ...systemProfile[0],
-        transforms: [wrappable],
-        props: { isStatic: true },
+        transforms: [sortable, wrappable],
       };
 
       tags = {
         ...tags[0],
       };
 
-      let columnList = [displayName, tags, systemProfile, lastSeenColumn];
+      groups = {
+        ...groups[0],
+        transforms: [wrappable],
+        props: { ...groups[0].props, isStatic: true }, // remove this line after group sorting is enabled in the API
+      };
+
+      let columnList = [
+        displayName,
+        groups,
+        tags,
+        systemProfile,
+        lastSeenColumn,
+      ];
 
       // Add column for impacted_date which is relevant for the rec system details table, but not pathways system table
       if (!pathway) {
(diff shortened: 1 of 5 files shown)


mhuth:
Well, the I see the codecov tests fail, but the files I have changed don't have any tests ATM anyway!

vkrizan:
@mhuth this would likely be conflicting with @mkholjuraev's changes. Can you please sync up? For the edge parity we'd like the changes to be made on backend first. Thank you!

mkholjuraev:
@vkrizan yes I agree that the backend needs modifications so that we can enable sorting and filtering for both edge and conventional systems. But, if we want to show the groups column ahead of backend changes, these changes in the PR do not conflict with my changes. I will suggest a few adjustments in this PR and then this will be a starting phase of the work.

mhuth:
It seems like an easy one to get out there in front of customers.  The FE changes in this PR work with the existing backend.  Another PR for sorting and filtering (for conventional systems) can be created once the API work is available.

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -307,15 +309,27 @@ const Inventory = ({
...
 
       tags = {
         ...tags[0],
       };
 
-      let columnList = [displayName, tags, systemProfile, lastSeenColumn];
+      groups = {
+        ...groups[0],
+        transforms: [wrappable],
+        props: { isStatic: true },
```
mkholjuraev, line 323: We need to make sure that column props are extended with `isStatic: true` instead of being mutated fully. Because there are other column props coming from Inventory to give the column the correct width, etc.

src/PresentationalComponents/Inventory/helpers.js
```diff
@@ -69,13 +69,11 @@ export const getEntities =
...
     } = config;
 
-    //operating_system is currently not supported, but will be down the line.
-    const sort =
-      orderBy === 'operating_system'
-        ? 'rhel_version'
-        : `${orderDirection === 'ASC' ? '' : '-'}${
-            orderBy === 'updated' ? 'last_seen' : orderBy
-          }`;
+    const sort = `${orderDirection === 'ASC' ? '' : '-'}${
```
mkholjuraev, line 72: Can we extract this into a utility function so that it can be reused and clean this function a bit?

mkholjuraev (commented):
@mhuth thank you. A few possible improvements and then we can proceed.

src/PresentationalComponents/Inventory/helpers.js
mhuth: Good idea.  Done!

src/PresentationalComponents/Inventory/Inventory.js
mhuth: Good catch, I didn't notice that!  Done!

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -302,20 +304,30 @@ const Inventory = ({
...
       };
 
-      let columnList = [displayName, tags, systemProfile, lastSeenColumn];
+      groups = {
+        ...groups[0],
+        transforms: [wrappable],
+        props: { ...groups[0].props, isStatic: true }, // remove this line after group sorting is enabled in the API
+      };
+
+      let columnList = [
```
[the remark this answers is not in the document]
mhuth: It sounds very challenging @gkarat :)  If there were existing tests on the Inventory* components I could use as a guide I'd give it a go.  Perhaps I can try it in a separate PR.  Seems there are lots of tests that are yet to be written for many components.

PR #1206, opened 2023-11-29
fix(RuleDetails): pass correct argument to viewAffectedSystems to fix intl warning

mkholjuraev:
# Description

fixes the intl warning that was caused by passing incorrect message value. As we are migrating away intl, I tried to replace this intl message with plain text. However I have found out that advisor-components FEC package, heavily relies on intl and we need to allocate dedicated effort to remove intl from both advisor, advisor4openshift and adv-components 

image

# How to test the PR

1. Run the PR
2. Go to recommendations table
3. Observe that the warning is gone

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -913,13 +913,6 @@ export default defineMessages({
     description: 'Error',
     defaultMessage: 'Error',
   },
-  viewAffectedSystems: {
-    id: 'viewAffectedSystems',
-    description:
-      'Link text to view all hosts that are affected by a recommendation',
-    defaultMessage:
-      'View {systemsCount, plural, one {the affected system} other {# affected systems}}',
-  },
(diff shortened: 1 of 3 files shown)


AsToNlele (approved):
LGTM thanks for the fix :smile: 

Could you please remove this console.log? :smile: 
https://github.com/RedHatInsights/insights-advisor-frontend/blob/653c6c7f9e136d3a80be497e55108ac105d2dec3/src/PresentationalComponents/ExecutiveReport/Download.js#L62

LiorKGOW:
Hey @mkholjuraev could you add the Jira ticket number please ?

mkholjuraev:
@LiorKGOW Hey, there was no associated ticket and thus the PR does not include it.

PR #1207, opened 2023-11-29
chore(fec): FEC package bump and updated wepback config

adonispuente:
This PR updates the FEC package to allow for the use of the updated web pack config settings. As more gets added there, more can be removed in the individual app.


bastilian (approved):
LGTM! @adonispuente!

PR #1213, opened 2023-12-06
fix(RHINENG-5907): remove disable rule kebab for pathway systems

mhuth:
https://issues.redhat.com/browse/RHINENG-5907

The toolbar kebab to disable rule(s) shouldn't be displayed for pathway systems because it doesn't make sense to disable individual pathway rules.  This simple PR / change removes the kebab when the inventory table shows pathway systems and displays it when they are individual rule systems.

- [X] The commit message has the Jira ticket linked

Before PR - pressing the kebab leads to a 404:
Screenshot from 2023-12-06 14-09-37

After PR - kebab removed:
Screenshot from 2023-12-06 13-53-27

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -391,6 +391,27 @@ const Inventory = ({
     },
   };
 
+  const getActionsConfig = () => {
+    const actions = [
+      <RemediationButton
+        key="remediation-button"
+        fallback={<Spinner size="md" />}
+        isDisabled={isRemediationButtonDisabled}
+        dataProvider={remediationDataProvider}
+        onRemediationCreated={(result) => onRemediationCreated(result)}
+      >
+        {intl.formatMessage(messages.remediate)}
+      </RemediationButton>,
+    ];
+    !pathway &&
+      actions.push({
+        label: intl.formatMessage(messages.disableRuleForSystems),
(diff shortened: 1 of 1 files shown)


mkholjuraev (approved):
LGTM!

mhuth:
@mkholjuraev can you review this one again because I had to make some change after [PR 1215](https://github.com/RedHatInsights/insights-advisor-frontend/pull/1215).  Thanks!

src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -391,6 +391,27 @@ const Inventory = ({
     },
   };
 
+  const getActionsConfig = () => {
```
mkholjuraev, line 394: Nice! It maybe even better to extract this util function out of the component into some helper file.

mhuth: Ok, I'll have a look into it, but there are a few values in there that are dependent / specific to the Inventory component.  It might be tricky to generalize the function.

mkholjuraev (approved):

src/PresentationalComponents/Inventory/Inventory.js
mkholjuraev: Aha, in that case, feel free to proceed please. Soon, we will release edge parity work and it is better to get this commit together.

PR #1164, opened 2023-08-17
Recommendations Page - Adding OverviewDashbar

LiorKGOW:
# Description

- Replaced PathwaysPanel with OverviewDashbar on Recommendations page
- Added tooltips to the titles and appropriate labeling or icons fit the cards
- Added filtering options when pressing on the numbers on the cards and switch to the appropriate tab
- Changed the title from "Advisor recommendations" to "Recommendations" to match the name of the tab on the side menu
- All added components are covered by tests in React Testing Library

Associated Jira ticket: [#ADVISOR-3035](https://issues.redhat.com/browse/ADVISOR-3035) 

# How to test the PR

Insure the functionality works, even when other filtering is already active

# Before the change
Recommendations Page: 

Recommendations Page - Before

# After the change
Recommendations page:

Whole page

Pressing on the number on Pathways card, would switch the view to Pathways tab:

Whole page - after pressing the Pathways the view would switch to the Pathways tab

Filtering by Incidents (by clicking on the number on the Incidents card), while also switching the active tab to be Recommendations tab:

![Whole page - after filtering by Incidents also the tab switches to Recommendations tab]
(https://github.com/RedHatInsights/insights-advisor-frontend/assets/93318917/7f47917c-6efd-4319-96c1-c237f2e8a144)

Filtering by Critical Recommendations:

Whole page - after filtering by Critical Recommendations

Filtering by Important Recommendations:

Whole page - after filtering by Important Recommendations

Pathways & Incidents cards contain tooltips with explanations:

Pathways & Incidents cards contain tooltips with explanations

Labels contain tootltips with explanations:

Labels contain tootltips with explanations

# Checklist:

- [x] Waiting for [Add the stats/overview/ endpoint for summary page information](https://gitlab.cee.redhat.com/insights-platform/advisor-backend/-/merge_requests/575) to be merged & tested
- [x] Fetch the card values from the api & present them on the cards
- [x] Create a test for the main component "OverviewDashbar"
- [x] Create a mock for the above test for the API endpoint
- [X] Add tests for the new components
- [X] Added screenshots to explain changes
- [ ] Rebase with the latest master branch
- [ ] re-open this PR with the correct branch

PR #1220, opened 2023-12-13
feat(RHINENG-5544): Remove Prefix From Breadcrumbs

LiorKGOW:
# Description

* This PR addresses the breadcrumbs issue that is mentioned on the ticket
* Also, removes a use of `useIntl` that we are trying to remove from this app

Associated Jira ticket: [#RHINENG-5544](https://issues.redhat.com/browse/RHINENG-5544)

# How to test the PR

Ensure the breadcrumbs does not show the app name

# Before the change

Recommendations:

Before: Recommendations

Systems:

Before: Systems

Topics: 

Before: Topics

# After the change

Recommendations:

After: Recommendations

Systems:

After: Systems

Topics:

After: Topics

### DIFF
--- a/src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
+++ b/src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
@@ -5,12 +5,10 @@ import { BreadcrumbItem } from '@patternfly/react-core/dist/esm/components/Bread
 import PropTypes from 'prop-types';
 import messages from '../../Messages';
 import { useGetRecQuery } from '../../Services/Recs';
-import { useIntl } from 'react-intl';
 import { useLocation } from 'react-router-dom';
 import Link from '@redhat-cloud-services/frontend-components/InsightsLink';
 
 const Breadcrumbs = ({ current }) => {
-  const intl = useIntl();
   const location = useLocation().pathname?.split('/');
   const [items, setItems] = useState([]);
   const skip =
@@ -24,11 +22,16 @@ const Breadcrumbs = ({ current }) => {
   useEffect(() => {
     const buildBreadcrumbs = () => {
       let crumbs = [];
+
       // add base
-      crumbs.push({
-        title: `${intl.formatMessage(messages.insightsHeader)} ${location[3]}`,
-        navigate: `/${location[3]}`,
-      });
+      if (location[3]) {
+        const baseNameWithCapitalLetter =
+          location[3].slice(0, 1).toUpperCase() + location[3].slice(1);
+        crumbs.push({
+          title: `${baseNameWithCapitalLetter}`,
+          navigate: `/${location[3]}`,
+        });
+      }
 
       // if applicable, add :id breadcrumb
       if (!skip) {
@@ -41,7 +44,7 @@ const Breadcrumbs = ({ current }) => {
       if (location[2] === 'pathways') {
         crumbs = [
           {
-            title: 'Advisor pathways',
+            title: 'Pathways',
             navigate: '/recommendations/pathways',
           },
         ];
@@ -66,7 +69,7 @@ const Breadcrumbs = ({ current }) => {
           <BreadcrumbItem isActive>{current}</BreadcrumbItem>
         </Breadcrumb>
       ) : (
-        intl.formatMessage(messages.loading)
+        messages.loading.defaultMessage
       )}
     </React.Fragment>
   );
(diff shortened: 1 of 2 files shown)


src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
```diff
@@ -66,7 +69,7 @@ const Breadcrumbs = ({ current }) => {
           <BreadcrumbItem isActive>{current}</BreadcrumbItem>
         </Breadcrumb>
       ) : (
-        intl.formatMessage(messages.loading)
+        messages.loading.defaultMessage
```
gkarat, line 72: We are making a soft deprecation of intl. This means, you no longer need to use the package and can freely insert a plain string "Loading..." here instead.

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
```diff
@@ -41,7 +44,7 @@ const Breadcrumbs = ({ current }) => {
       if (location[2] === 'pathways') {
         crumbs = [
           {
-            title: 'Advisor pathways',
+            title: 'Pathways',
             navigate: '/recommendations/pathways',
           },
```
gkarat, line 49: I am not really sure this works as expected: try to navigate to any pathways details page and you will see Recommendations as the first breadcrumb. Shouldn't it be Pathways instead and lead to /recommendations/pathways instead of /recommendations?

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
gkarat, line 1: TBH, I don't really like the location parts shuffling. Can we consider something like this https://reactrouter.com/en/main/hooks/use-matches and rewrite this breadcrumbs logic?

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
LiorKGOW: Isn't it what I am doing here ? 😅 
I removed the intl line and inserted a use of `messages` object, which inserts a raw string 

This is the `loading` attribute of messages:
loading

gkarat: I mean, you don't even need to use `messages` because it's already a part of react-intl API (see what is exported in Messages.js) 🙂

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
gkarat: Feel free to skip this ^ it's not much related to the PR and can be of a larger effort

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
LiorKGOW: I can see what you mean, but I thought we were trying to avoid using react-intl

@gkarat please have a look at this comment:
https://github.com/RedHatInsights/insights-advisor-frontend/pull/1164#issuecomment-1782566517 
That was the reason behind this line

Just wanted to make sure we are all on the same page here before updating this line @gkarat @Fewwy

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
LiorKGOW: I can see why you think it should be removed @gkarat 
I can open an issue about refactoring this component, and in the refactor we can change the logic to be with `useMatches` instead of `useLocation`

What do you think @gkarat ?

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
LiorKGOW: This issue is fixed in the second commit I added 👍🏼

LiorKGOW:
Hey @gkarat I have went over all of your comments and fixed the bug you have mentioned.
This PR is ready for a re-review :)

src/PresentationalComponents/Breadcrumbs/Breadcrumbs.js
gkarat: My whole point is about using `"Loading..."` here directly instead of `messages.loading.defaultMessage`:
```suggestion
        Loading...
```

In my previous comment, I was trying to point out that if you use the `messages` variable from the Messages.js module, you utilise intl automatically. We want to avoid this, and this is why I suggest to replace with just a plain text (messages.loading.defaultMessage is not).

gkarat (approved):
LGTM! Thank you @LiorKGOW

PR #1221, opened 2023-12-18
refactor(RHINENG-6162): Refactor pdf generator

adonispuente:
In an attempt to diagnose the pdf generator issue ,this PR refactors bits of the Download.js file : https://issues.redhat.com/browse/RHINENG-6162

### DIFF
--- a/src/PresentationalComponents/ExecutiveReport/Download.js
+++ b/src/PresentationalComponents/ExecutiveReport/Download.js
@@ -15,6 +15,7 @@ import { Get } from '../../Utilities/Api';
 import messages from '../../Messages';
 import { useIntl } from 'react-intl';
 import { useDispatch } from 'react-redux';
+import { BuildExecReport } from './BuildExecReport';
 
 const DownloadExecReport = ({ isDisabled }) => {
   const intl = useIntl();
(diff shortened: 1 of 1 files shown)


src/PresentationalComponents/ExecutiveReport/Download.js
```diff
@@ -23,30 +24,29 @@ const DownloadExecReport = ({ isDisabled }) => {
...
+      ).map(({ data }) => data);
 
-      const report = BuildExecReport({
-        statsReports,
-        statsSystems,
-        topActiveRec,
-        intl,
-      });
+      const report = (
+        <BuildExecReport
```
mkholjuraev, line 43: I believe that the report should not be pre-rendered when it is passed to the generator service. For example, [vulnerability](https://github.com/RedHatInsights/vulnerability-ui/blob/5e9c2cce07773dc95fea302bcd8b94b9a08233ea/src/Components/SmartComponents/Reports/DownloadSystemsReport.js#L85) also does not return pre-rendered component.

bastilian: What do you mean with "pre-rendered"?

bastilian: If you mean call the `BuildExecReport` component (properly) as a component then, no it should not be called as a function. Yes, it works and Vulnerability is doing it too this way, but it is not correct and can interfere with Reacts correct behaviour. 

If it returns JSX it is a component, and components should be called as such.

mkholjuraev: ah, I got confused as I was looking at vulnerability while reviewing

bastilian (approved):
Lets get this merged! Thank you @adonispuente!

bastilian:
@adonispuente For now we can ignore the coverage warnings, I think, but we should eventually properly extract things into hooks and have tests for those. In case this does not help with the download we should look into that next as the tests could also give a hint at what goes wrong.

PR #1231, opened 2024-01-15
fix: URL encode the pipe character in rule ids

mhuth:
# Description
This PR URL encodes RHEL Advisor recommendation / rule id values whenever they are sent to the backend so the request may pass through environments enforcing strict HTTP standards (eg not allowing the pipe character "|" in URLs)

This came from a Slack discussion and I'm not sure if a Jira has been created.

Before the change, the pipe character is sent in URLs unencoded:
Screenshot from 2024-01-15 13-16-15

After the change, the pipe character is encoded:
Screenshot from 2024-01-15 13-18-17

# Checklist:

### DIFF
--- a/src/SmartComponents/Recs/helpers.js
+++ b/src/SmartComponents/Recs/helpers.js
@@ -18,7 +18,7 @@ export const enableRule = async (
   handleModalToggle
 ) => {
   try {
-    await DeleteApi(`${BASE_URL}/ack/${rule.rule_id}/`);
+    await DeleteApi(`${BASE_URL}/ack/${encodeURI(rule.rule_id)}/`);
     addNotification({
       variant: 'success',
       timeout: true,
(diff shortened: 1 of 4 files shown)


gkarat (approved):
LGTM! Thanks @mhuth

src/SmartComponents/Recs/helpers.js
```diff
@@ -76,8 +80,12 @@ export const bulkHostActions = async ({
...
-      conventionalURL: `/api/insights/v1/rule/${ruleId}/systems_detail/?filter[system_profile][host_type][nil]=true&limit=1`,
-      edgeURL: `/api/insights/v1/rule/${ruleId}/systems_detail/?filter[system_profile][host_type]=edge&limit=1`,
+      conventionalURL: `/api/insights/v1/rule/${encodeURI(
+        ruleId
+      )}/systems_detail/?filter[system_profile][host_type][nil]=true&limit=1`,
+      edgeURL: `/api/insights/v1/rule/${encodeURI(
+        ruleId
+      )}/systems_detail/?filter[system_profile][host_type]=edge&limit=1`,
     };
   }
```
lphiri, line 91: @mhuth  "[" is also an unsafe character per RFC .https://www.ietf.org/rfc/rfc1738.txt.   Can we encodeURI the whole URL (segment) to catch all of these?

lphiri: cc @Hyperkid123  - they may be a way to handle this globally in UI framework.

Hyperkid123: @lphiri Yes we can do this globaly. We just have to experiment a little bit to see what happens if we try to encode already encoded URI. Any changes to the global patch have obviously global impact and we have to be very careful.

PR #1232, opened 2024-01-16
Update patternfly react-tokens

aneelac22:
[RHINENG-6065](https://issues.redhat.com/browse/RHINENG-6065)

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "@patternfly/react-core": "^4.276.6",
     "@patternfly/react-icons": "^4.93.6",
     "@patternfly/react-table": "^4.113.0",
-    "@patternfly/react-tokens": "^4.94.6",
+    "@patternfly/react-tokens": "^5.1.2",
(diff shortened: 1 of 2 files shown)


dkuc (approved):

package.json
```diff
@@ -73,7 +73,7 @@
     "@patternfly/react-core": "^4.276.6",
     "@patternfly/react-icons": "^4.93.6",
     "@patternfly/react-table": "^4.113.0",
-    "@patternfly/react-tokens": "^4.94.6",
+    "@patternfly/react-tokens": "^5.1.2",
```
bastilian, line 76: The v5 package may conflict with the v4 packages and we should refrain from mixing them.

bastilian (changes requested):

aneelac22:
I am actually closing this, I was trying to solve a specific chunk load error we were seeing in Fedramp but, we ended up reverting it back to when it was working to avoid instability for release tomorrow.

PR #1233, opened 2024-01-30
chore(RHINENG-3609/RHINENG-5927): PF5 and react18 migration

adonispuente:
React 18 and pf5 upgrade.
R18: [RHINENG-3609](https://issues.redhat.com/browse/RHINENG-3609)
PF5: [RHINENG-5927](https://issues.redhat.com/browse/RHINENG-5927)


src/PresentationalComponents/RulesTable/helpers.js
```diff
@@ -222,7 +224,7 @@ export const filterConfigItems = (
     },
     {
       label: FC.total_risk.title,
-      type: FC.total_risk.type,
+      type: conditionalFilterType.text,
```
bastilian, line 227: We need to make sure these are correct. `FC.total_risk.type` returns "checkbox", this here would return `text`. https://github.com/RedHatInsights/insights-advisor-frontend/blob/master/src/AppConstants.js#L199

adonispuente: I'll double check now and fix these. I saw that the type was a string, so I had just replaced that with text.

adonispuente: good catch! I've sent up the updated values

bastilian (approved):
Nice work! Thank you @adonispuente!

Looks all good to me! There are some minor visual differences, but I believe these are intentional in some cases and in others do actually fix issues.

PR #1234, opened 2024-02-07
feat(RHINENG-3078): extend systems filter with new hybrid options

mkholjuraev:
…o maintain

# Description

Associated Jira ticket: RHINENG-3078

This PR extends the systems impacted filter in advisor to include the new options dnfyum, ostree for only accounts with edge systems. For the rest of the users, behavior should stay the same. After trying out a bunch of approaches, I decided to have simple if else conditions to map the filter values to the complex API spec in `mapUpdateMethodFilterToAPISpec` func.

API filter syntax [(RHINENG-5353)](https://issues.redhat.com/browse/RHINENG-5353)

- All rules: no query parameter
- No impacted systems: impacting=false.
- Only conventional systems: impacting=true&update_method=dnfyum.
- Only immutable systems: impacting=true&update_method=ostree.
- Any system: impacting=true.

I wrote new tests to cover the work, but I had to comment them out as most of the tests are commented out due to the PF migrations :))))

This also cleans up the RulesTable component a bit by moving helpers functions into a separate file. 

# How to test the PR

1. Run the app
2. navigate to recommendations table in landing page
3. Observe that now the table is filtered with both dnfyum and ostree and the filter chips are present
4. Play with the impacting filter with new options, observe it works as expeted. 
5. Repeat the same actions in all pages where rules table is present 

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/PresentationalComponents/Common/Tables.js
+++ b/src/PresentationalComponents/Common/Tables.js
@@ -49,9 +49,42 @@ export const buildTagFilter = (tagFilters) => {
   };
 };
 
+const mapUpdateMethodFilterToAPISpec = (filters) => {
+  //If none option is chosen
+  if (filters?.update_method?.includes('none')) {
+    //remove update_method filter from API options and set impacting to false if only none options is chosen
+    if (filters.update_method === 'none') {
+      delete filters.update_method;
+      filters.impacting = 'false';
+
+      //in any other cases, remove 'none' option from the API options as it does have any handler
+      //concatenate or set false to impacting API option
+    } else {
+      const filteredValues = filters.update_method.replace(
+        /,?none||none,?/g,
+        ''
+      );
+
+      filters.update_method = filteredValues;
+      filters.impacting =
+        filters.impacting.toString().length > 1
+          ? filters.impacting.concat(',false')
+          : 'false';
+    }
+  }
+
+  //when user deselects all update_method filters remove the both update_method and impacting filters
+  if (filters?.update_method === '') {
+    delete filters.update_method;
+    delete filters.impacting;
+  }
+
+  return filters;
+};
+
 // transforms array of strings -> comma seperated strings, required by advisor api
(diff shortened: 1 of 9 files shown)


src/PresentationalComponents/Common/Tables.js
```diff
@@ -49,9 +49,42 @@ export const buildTagFilter = (tagFilters) => {
...
+          ? filters.impacting.concat(',false')
+          : 'false';
+    }
+  }
+
+  //when user deselects all update_method filters remove the both update_method and impacting filters
+  if (filters?.update_method === '') {
+    delete filters.update_method;
+    delete filters.impacting;
+  }
```
AsToNlele, line 80: I'd put it in an `else if` instead of an `if`.
If the `if` condition above this one would get refactored and the update_method changed to `''` it could trigger this condition and it looks separated this way

AsToNlele:
Just to double check, when the filters are like this
image
The query is like this: `impacting=true&&update_method=ostree%2Cdnfyum` is it correct?

AsToNlele:
I have a brand new account and uploaded some systems (some `edge` as well) I can see them in Advisor but not in Inventory. 
image
The api endpoint on Inventory checks on this endpoint `/api/edge/v1/devices/devicesview` where I get 0 devices but here we check on this endpoint `/api/inventory/v1/hosts?filter[system_profile][host_type]=edge` and I get 2 devices. Do you think this is worth reporting or not? I'm not sure, because I guess people register it the correct way like in the picture above

src/PresentationalComponents/Common/Tables.js
mkholjuraev: hmm, this makes sense. Will add it. Thank you!

AsToNlele (approved):
Thanks for fixing the suggestions, LGTM :smile: 
Nice PR number btw :smile:

PR #1241, opened 2024-03-05
fix(RHINENG-3078): remove update_method and impacting filters when al…

mkholjuraev:
…l system fitler types are chosen

# Description

Associated Jira ticket: # https://issues.redhat.com/browse/RHINENG-3078

Removes both impacting and update_method filters when all system filter types are chosen.

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/PresentationalComponents/Common/Tables.js
+++ b/src/PresentationalComponents/Common/Tables.js
@@ -1,5 +1,6 @@
 import { generateFilter } from '@redhat-cloud-services/frontend-components-utilities/helpers';
 import { SYSTEM_FILTER_CATEGORIES, SYSTEM_TYPES } from '../../AppConstants';
+import isEqual from 'lodash/isEqual';
 
 // Builds returns url params from table filters, pushes to url if history object is passed
 export const urlBuilder = (filters = {}) => {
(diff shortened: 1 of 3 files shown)


src/PresentationalComponents/Common/Tables.js
```diff
@@ -49,9 +50,19 @@ export const buildTagFilter = (tagFilters) => {
...
+  if (
+    filters?.update_method === '' ||
+    isEqual(filters?.update_method?.split(',').sort(), [
+      'dnfyum',
+      'none',
+      'ostree',
+    ])
+  ) {
+    //when user deselects all update_method filters remove the both update_method and impacting filters
+    delete filters.update_method;
```
bastilian, line 63: I think we should prefer returning a new object rather than mutating `filters`. We have seen these `delete`s cause issues and unexpected states/results.

package-lock.json
bastilian, line 1: Is this intended?

src/PresentationalComponents/Common/Tables.js
mkholjuraev: Yup, actually there is a new object created to return [here](https://github.com/RedHatInsights/insights-advisor-frontend/blob/2e2897c9889e78a047ed9284f89e1029273cfb6a/src/PresentationalComponents/Common/Tables.js#L89). Should I still create a new filter return here as well? Happy to do the best approach, let me know

package-lock.json
mkholjuraev: Done!

computercamplove:
PR is tested locally - works as expected

src/PresentationalComponents/Common/Tables.js
Fewwy: I think Sebastian is correct. You already importing lodash in this file, so you can use something like deepClone method to avoid changing the initial filters object
[_.deepClone](https://www.geeksforgeeks.org/lodash-_-clonedeep-method/)

mkholjuraev: @Fewwy I have made the change, please take a look!

Fewwy (approved):

src/PresentationalComponents/Common/Tables.js
bastilian: @mkholjuraev @Fewwy It would have also been good to completely avoid the `delete` and not need to clone the object. Using spreading and destructing the same can be achieved and it creates a new object, without cloning. See for example: https://javascript.plainenglish.io/avoid-the-delete-keyword-in-javascript-87ff2a47f26c

mkholjuraev: I will keep this in mind and will do the required change as soon as I have  a chance

PR #1253, opened 2024-04-09
Block remediations for bootc systems

dkuc:
https://issues.redhat.com/browse/RHINENG-9195

I couldn't figure out how to get the system data backout of the InventoryTable so I wrote a wrapper of the api call that captures the data.

image


dkuc:
Closing for now to avoid confusion

PR #1254, opened 2024-04-11
fix(RHINENG-8460): refactor of edge&conventional devices queries and fix for filters

Fewwy:
# Description
There was a race condition in the table caused by the fact that we were creating filter chips too early without waiting for the edge devices

https://issues.redhat.com/browse/RHINENG-8460

What I did in this PR:
I refactored the way how we request the data about edge and conventional systems. I removed try catch and now we are using the RTK query hook and passing it through the context.
Now the filter builder is requesting the data based on the edge and conventional systems requests that are resolved.
The number of requests should stay the same way as before - we request 2 endpoints for data on the edge and conventional devices, and then based on the presence/absence of these devices we make filters.
I fixed the tests for the Rules table because they were commented out.
I updated Cypress Library because the new version is faster.

How to test:
There are different scenarios:
Scenario 1: Open the table, add a filter, copy the URL, and open it in a new tab - the filter should be present, chips should be generated and filtered results should be the same
Scenario 2: Open the table from the dashboard links - the table should be opened with the filter pre-applied.
Scenario 3: Force edge devices to be false/true and repeat scenarios 1 and 2


src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -160,12 +160,15 @@ const RulesTable = ({ isTabActive }) => {
   }, []);
 
   useEffect(() => {
-    dispatch(
-      updateRecFilters({
-        ...filtersInitialState.recState,
-        ...getDefaultImpactingFilter(hasEdgeDevices),
-      })
-    );
+    if (!isLoading) {
```
bastilian, line 163: The `isLoading` should probably be a dependency of this effect, or?

AsToNlele (dismissed):
works, but I think this can be somehow improved

gkarat:
I think Sebastian's comment makes sense. Can we add isLoading to useEffect's deps before merging?

src/PresentationalComponents/RulesTable/RulesTable.js
mkholjuraev: I tried running with isLoading added into dependencies. It seems if I add it, the bug still exists. But, if I remove it, the bug is gone.

bastilian: What is this `useEffect` trying to do? 

If the `isLoading` is used it should be a dependency it not working correctly if it is a dependency might point to a general issue with the flow in the component states.

LiorKGOW:
Hey @Fewwy 
I saw this PR contains a 'bug' label
Does it still have a bug ? Or should we remove it ?

Fewwy:
@LiorKGOW Check the ticket, there was a bug

LiorKGOW (dismissed):
Hey @Fewwy 
Had a few comments, please let me know what you think :)

src/ZeroStateWrapper.js
```diff
@@ -21,34 +24,39 @@ export const ZeroStateWrapper = ({ children }) => {
...
+  } = useGetEdgeDevicesQuery();
+  const {
+    data: conventional,
+    isSuccess: conventionalQuerySuccess,
+    isError: conventionalError,
+    error: conventErrorMessage,
+  } = useGetConventionalDevicesQuery();
+  useEffect(() => {
+    setHasEdgeDevices(edge?.total > 0 ? true : false);
+    setHasConventionalSystems(conventional?.total > 0 ? true : false);
```
gkarat, line 41:
```suggestion
    setHasEdgeDevices(edge.total > 0);
    setHasConventionalSystems(conventional.total > 0);
```

gkarat (approved):
LGTM. It looks like the PR fixes the bug, I couldn't spot any issues when working with filters/refreshing the page. I have left only one comment on the code. The solution with context looks neat. https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequerystate could possible work here as well. Thanks @Fewwy

PR #1257, opened 2024-04-16
Create webpack.yml for sentry config

adonispuente:
Adds sentry configs

### DIFF
--- a/.github/workflows/webpack.yml
+++ b/.github/workflows/webpack.yml
@@ -21,3 +21,4 @@ jobs:
           tagName: ${{ github.ref }}
           environment: master
           releaseNamePrefix: Rhel-Advisor
+          sourcemaps: 'dist/sourcemaps'
--- a/.github/workflows/webpack.yml
+++ b/.github/workflows/webpack.yml
@@ -1,8 +1,9 @@
 name: sentryConfig
 
 on:
-  release:
-    types: [published, prereleased]
+  push:
+    branches:
+      - 'master'
 
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/webpack.yml
```diff
@@ -0,0 +1,24 @@
+name: sentryConfig
+
+on:
+  release:
+    types: [published, prereleased]
```
gkarat, line 5: Are we going to create release tags in order to trigger the workflow? Otherwise, the workflow won't be started according to the documentation.

Didn't we want to update source maps every time we push to master? Like this
```
on:
  push:
    branches:
      - 'master'
```

.github/workflows/webpack.yml
```diff
@@ -0,0 +1,24 @@
...
+
+
+jobs:
+  createSentryRelease:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
+      - name: Create a Sentry.io release
+        uses: tclindner/sentry-releases-action@v1.0.0
```
gkarat, line 15: I would vote for using an official action instead: https://github.com/marketplace/actions/sentry-release

.github/workflows/webpack.yml
```diff
@@ -0,0 +1,24 @@
...
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
+        with:
+          tagName: ${{ github.ref }}
+          environment: master
+          releaseNamePrefix: Rhel-Advisor
```
gkarat, line 23: Maybe we better use the prefixes that are up-to-date with chrome-service-backend? Like in our case `advisor` https://github.com/RedHatInsights/chrome-service-backend/blob/main/static/stable/prod/modules/fed-modules.json#L224

adonispuente: im not opposed to this, I mainly just wanted to differenciate ocp-advisor and rhel-advisor down the line

.github/workflows/webpack.yml
adonispuente: sending this up now

gkarat (approved):

PR #1258, opened 2024-04-16
feat(sentryConfig): Adds sentry configuration

adonispuente:
Adds configuration to add source maps to sentry

### DIFF
--- a/config/prod.webpack.config.js
+++ b/config/prod.webpack.config.js
@@ -25,24 +25,27 @@ plugins.push(
         './RootApp': resolve(__dirname, '../src/AppEntry'),
         './SystemDetail': resolve(__dirname, '../src/Modules/SystemDetail'),
       },
-    },
-    ...(process.env.SENTRY_AUTH_TOKEN
-      ? sentryWebpackPlugin({
-          authToken: process.env.SENTRY_AUTH_TOKEN,
-          org: process.env.SENTRY_ORG,
-          project: process.env.SENTRY_PPROJECT,
-        })
-      : {})
+    }
   )
 );
 
-const SourceMapsPlugin = new webpack.SourceMapDevToolPlugin({
-  test: /\.js/i,
-  exclude: /(node_modules|bower_components)/i,
-  filename: `sourcemaps/[name].js.map`,
-});
+if (process.env.SENTRY_AUTH_TOKEN) {
+  plugins.push(
+    sentryWebpackPlugin({
+      authToken: process.env.SENTRY_AUTH_TOKEN,
+      org: process.env.SENTRY_ORG,
+      project: process.env.SENTRY_PROJECT,
+    })
+  );
 
-plugins.push(SourceMapsPlugin);
+  const SourceMapsPlugin = new webpack.SourceMapDevToolPlugin({
+    test: /\.js/i,
+    exclude: /(node_modules|bower_components)/i,
+    filename: `sourcemaps/[name].js.map`,
+  });
+
+  plugins.push(SourceMapsPlugin);
+}
 
 module.exports = () => {
   const CopyFilesWebpackPlugin = new (require('copy-webpack-plugin'))({
@@ -54,8 +57,11 @@ module.exports = () => {
       },
     ],
   });
-  'source-map', // Source map generation must be turned on
-    plugins.push(CopyFilesWebpackPlugin);
+  plugins.push(CopyFilesWebpackPlugin);
 
-  return { ...webpackConfig, plugins };
+  return {
+    ...webpackConfig,
+    ...(process.env.SENTRY_AUTH_TOKEN ? { devtool: 'source-map' } : {}), // Source map generation must be turned on
+    plugins,
+  };
 };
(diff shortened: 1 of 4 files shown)


config/prod.webpack.config.js
```diff
@@ -23,10 +25,23 @@ plugins.push(
         './RootApp': resolve(__dirname, '../src/AppEntry'),
         './SystemDetail': resolve(__dirname, '../src/Modules/SystemDetail'),
       },
-    }
+    },
+    sentryWebpackPlugin({
```
bastilian, line 29: We should maybe only configure the plugin if there is a `process.env.SENTRY_AUTH_TOKEN` set. It might also be good to put the `org` and `project` in an environment variable as well.

adonispuente: Actually I disagree I dont think it should be an env variable. This should always be a constant variable, it should not be manipulated.

bastilian: ... Please use the env vars the GH action will receive. :) https://github.com/RedHatInsights/insights-advisor-frontend/pull/1257/files#diff-c39f042a9299790ccb1468277471730df5efdfa8105f30b035c8554a5b00386eR17-R19

bastilian: > Maybe my understanding is incorrect but the variables definitely shouldnt be revealed.

Maybe there was a misunderstanding. In the other PR the "secrets" are passed as (encrypted) env variables. To not put things like that directly in a file and on GitHub or expose them via the GH action logs.

adonispuente: I see, i understand now. updating the PR

adonispuente: the ci test fail repeatedly with these changes. 

15:46:46 [webpack-cli] Failed to load '/container_workspace/config/prod.webpack.config.js' config

bastilian: What was the error why it was not able to load the config?

adonispuente: https://ci.int.devshift.net/job/RedHatInsights-insights-advisor-frontend-pr-check/1133/console
These are the logs for that commit.

Its upset with the iterator. I wasnt sure why with the conditional set up, but it doesnt accept it. 



`12:10:17 [webpack-cli] Failed to load '/container_workspace/config/prod.webpack.config.js' config
12:10:17 [webpack-cli] TypeError: Found non-callable @@iterator
12:10:17     at Object.<anonymous> (/container_workspace/config/prod.webpack.config.js:16:81)
12:10:17     at Module._compile (node:internal/modules/cjs/loader:1198:14)
12:10:17     at Object.Module._extensions..js (node:internal/modules/cjs/loader:1252:10)
12:10:17     at Module.load (node:internal/modules/cjs/loader:1076:32)
12:10:17     at Function.Module._load (node:internal/modules/cjs/loader:911:12)
12:10:17     at Module.require (node:internal/modules/cjs/loader:1100:19)
12:10:17     at require (node:internal/modules/cjs/helpers:119:18)
12:10:17     at WebpackCLI.tryRequireThenImport (/container_workspace/node_modules/webpack-cli/lib/webpack-cli.js:204:22)
12:10:17     at loadConfigByPath (/container_workspace/node_modules/webpack-cli/lib/webpack-cli.js:1404:38)
12:10:17     at /container_workspace/node_modules/webpack-cli/lib/webpack-cli.js:1454:88
12:10:17 ERROR: "build" exited with 2.`

bastilian: Oh, we probably need to configure these env variables for CI too, However, if we have `process.env.SENTRY_AUTH_TOKEN  ? ... ` it shouldn't cause any issue. also, Do double check what `sentryWebpackPlugin` returns and if it can be spread into an object.

config/prod.webpack.config.js
```diff
@@ -23,10 +25,23 @@ plugins.push(
...
+      org: process.env.SENTRY_ORG,
+      project: process.env.SENTRY_PROJECT,
+    })
   )
 );
 
+const SourceMapsPlugin = new webpack.SourceMapDevToolPlugin({
+  test: /\.js/i,
+  exclude: /(node_modules|bower_components)/i,
+  filename: `sourcemaps/[name].js.map`,
```
bastilian, line 40: The assets generated by the shared webpack config will use the `contenthash`. It might be important for these to match up.
https://github.com/RedHatInsights/frontend-components/blob/0e90e6071e327881def444388907a83a3d5a1b1c/packages/config/src/lib/createConfig.ts#L99

adonispuente: when running everything is building out properly. I believe this is good.

config/prod.webpack.config.js
```diff
@@ -27,6 +29,24 @@ plugins.push(
...
+    })
+  );
+
+  const SourceMapsPlugin = new webpack.SourceMapDevToolPlugin({
+    test: /\.js/i,
+    exclude: /(node_modules|bower_components)/i,
+    filename: `sourcemaps/[name].js.map`,
+  });
+
+  plugins.push(SourceMapsPlugin);
```
gkarat, line 47: I think this plugin is not needed. Sentry Webpack plugin already generate the source maps, isn't it? I have tried to remove this plugin and check the /dist folder and could see `sourcemaps` folder created successfully.

gkarat (approved):

PR #1259, opened 2024-04-23
fix(RHINENG-9594): Only retrieve/display recs for the chosen pathway

mhuth:
https://issues.redhat.com/browse/RHINENG-9594

When selecting a pathway, the rules table component displays all rules regardless of whether they are associated with the selected pathway or not.

- Log into console.stage.redhat.com
- Go to the Insights -> Operations -> Advisor
- Select the '5' under Pathways - this lists the pathways.
- Choose one of the pathways - e.g. 'Update Kernel Boot Options'.
... all rules matching the filter conditions are displayed, not just the rules for 'Update Kernel Boot Options'

- [X] The commit message has the Jira ticket linked

### DIFF
--- a/src/PresentationalComponents/RulesTable/RulesTable.js
+++ b/src/PresentationalComponents/RulesTable/RulesTable.js
@@ -52,7 +52,7 @@ import { useActionsResolver } from './useActionsResolver';
 import { AccountStatContext } from '../../ZeroStateWrapper';
 import impactingFilter from '../Filters/impactingFilter';
 
-const RulesTable = ({ isTabActive }) => {
+const RulesTable = ({ isTabActive, pathway }) => {
   const intl = useIntl();
   const dispatch = useDispatch();
   const { search } = useLocation();
@@ -84,14 +84,11 @@ const RulesTable = ({ isTabActive }) => {
 
   const setFilters = (filters) => dispatch(updateRecFilters(filters));
 
-  let options = {};
-  selectedTags?.length &&
-    (options = {
-      ...options,
-      ...{ tags: selectedTags.join(',') },
-    });
-  workloads &&
-    (options = { ...options, ...workloadQueryBuilder(workloads, SID) });
+  const options = {
+    ...(selectedTags?.length ? { tags: selectedTags.join(',') } : {}),
+    ...(workloads ? workloadQueryBuilder(workloads, SID) : {}),
+    ...(pathway ? { pathway } : {}),
+  };
 
   const {
     data: rules = [],
@@ -341,6 +338,7 @@ const RulesTable = ({ isTabActive }) => {
 
 RulesTable.propTypes = {
   isTabActive: PropTypes.bool,
+  pathway: PropTypes.string,
 };
 
 export default RulesTable;
--- a/package-lock.json
+++ b/package-lock.json
@@ -4627,9 +4627,9 @@
       }
     },
     "node_modules/@patternfly/react-styles": {
-      "version": "5.2.1",
-      "resolved": "https://registry.npmjs.org/@patternfly/react-styles/-/react-styles-5.2.1.tgz",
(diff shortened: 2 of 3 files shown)


mhuth:
Before PR:
Screenshot from 2024-04-23 16-40-06

After PR:
Screenshot from 2024-04-23 16-30-58

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -92,6 +92,7 @@ const RulesTable = ({ isTabActive }) => {
     });
   workloads &&
     (options = { ...options, ...workloadQueryBuilder(workloads, SID) });
+  pathway && (options = { ...options, pathway });
```
leSamo, line 95: I'm not sure about the opinion of other FE engineers, but I find using short-circuiting inside LHS of assignments very hard to read. Don't get me wrong, I like concise code, but not at the cost of readability. I'd personally prefer if this assignment was either wrapped in `if` or the `&&` was inside RHS of the assignment. What do you think?

mhuth: I don't mind changing it, but I was just keeping with the style of the surrounding code.

leSamo: Oki, this might be something for a separate PR.

leSamo (dismissed):
LGTM!

src/PresentationalComponents/RulesTable/RulesTable.js
mhuth: If I change this line, then the lines above should be changed as well I guess?  To make it consistent?  But this assignment style for options has been used throughout the project, not just in this file.  If we change it here then should we change it in the other files too?  I vote we keep it consistent with existing code.  Its readable enough.

bastilian: I agree with you both. :D 

We should keep established patterns, where they exist, but the assignments here are hard to read done this way. A better and more common through out all apps would be something like:

```js
const options = {
   ...selectedTags?.length && { tags: selectedTags.join(',') } : {},
   ...workloads && workloadQueryBuilder(workloads, SID) : {},
   ...pathway && { pathway } : {}
};
```

<sup>*</sup> This is an example, please do double check before applying.

leSamo: @bastilian yes, this format would be perfect, I think you have redundant `: {}` at the end of each line

mhuth: I'll leave it as is, and go through the project at another time and change all the option assignments.  It could probably be put into some sort of library function.

However, I was curious about the failing checks.  Should this be addressed before merging the PR?  I'm not sure why the linter tests are failing though, it seems be something outside of this PR that's causing it to fail.  Your thoughts @leSamo @bastilian ?

leSamo: Yea, the deps are failing, I cannot reproduce the error locally :thinking:

bastilian:
@mhuth @leSamo We should fix the dependency issues before merging this in a separate PR and not risk having master fail then too.

The issue can also be reproduced locally, by running `npm ci`.

src/PresentationalComponents/RulesTable/RulesTable.js
```diff
@@ -52,7 +52,7 @@ import { useActionsResolver } from './useActionsResolver';
 import { AccountStatContext } from '../../ZeroStateWrapper';
 import impactingFilter from '../Filters/impactingFilter';
 
-const RulesTable = ({ isTabActive }) => {
+const RulesTable = ({ isTabActive, pathway = '' }) => {
```
bastilian, line 55: This will cause the `pathway` to always end up in the options, and I'm not sure if we want that.

It think if we make the lines bellow in the options something like `...(CONST_FOR_OPTIONS ? { CONST_FOR_OPTIONS }) : {},` we should not need to set a default for `pathway`, or workloads and selectedTags.

mhuth: Done!

mhuth: @bastilian linter and tests pass now.  Ok to merge it?

PR #1263, opened 2024-05-13
Change system counts inside tables to links

leSamo:
# Description

Associated Jira ticket: [RHINENG-3595](https://issues.redhat.com/browse/RHINENG-3595)

How to test:
1. Open recommendations table and click on system count -- you should be navigated to the recommendation detail page
2. Open pathways table and click on system count -- you should be navigated to the pathway detail page
3. Open topics table and click on system count -- you should be navigated to the topic detail page


gkarat (commented):
Can we have a few unit tests updated/added to ensure the columns render links instead of plain text with the expected pathnames?

leSamo:
@gkarat yes, good idea, I added the tests

gkarat (approved):

PR #1265, opened 2024-05-14
Update webpack.yml to build before sending to sentry

adonispuente:
Just builds the application before sending anything to sentry


AsToNlele (approved):
lets goo

PR #1267, opened 2024-05-16
fix(Redux): redux out of useMemo, remove redux logger and devtools in…

mkholjuraev:
… prod

# Description

Fixes the stage environment by removing the redux logger middleware. It also refactors the redux initialization out of useMemo.

# How to test the PR

# Before the change

# After the change

# Dependent work link

# Checklist:

### DIFF
--- a/src/Store/index.js
+++ b/src/Store/index.js
@@ -7,14 +7,11 @@ import { Recs } from '../Services/Recs';
 import { Systems } from '../Services/Systems';
 import { Topics } from '../Services/Topics';
 import filters from '../Services/Filters';
-import logger from 'redux-logger';
 import { notificationsMiddleware } from '@redhat-cloud-services/frontend-components-notifications/';
 import { notificationsReducer } from '@redhat-cloud-services/frontend-components-notifications/redux';
 import promiseMiddleware from 'redux-promise-middleware';
 import { SystemVariety } from '../Services/SystemVariety';
 
-const env = 'development';
-const production = env !== 'production';
 const reducer = {
   [Pathways.reducerPath]: Pathways.reducer,
   [Recs.reducerPath]: Recs.reducer,
@@ -52,17 +49,13 @@ const middleware = (getDefaultMiddleware) =>
     notificationsMiddleware({
       errorTitleKey: ['message'],
       errorDescriptionKey: ['response.data.detail'],
-    }),
-    production && logger
+    })
   );
 
-const getStore = () => {
-  return configureStore({
-    reducer,
-    middleware,
-    devTools: production,
-  });
-};
+const store = configureStore({
+  reducer,
+  middleware,
+});
 
 const updateReducers = (newReducers = {}) =>
   combineReducers({
@@ -70,4 +63,4 @@ const updateReducers = (newReducers = {}) =>
     ...newReducers,
   });
 
-export { getStore, updateReducers };
+export { store, updateReducers };
--- a/src/Store/index.js
+++ b/src/Store/index.js
@@ -25,20 +25,8 @@ const reducer = {
   entitiesDetailsReducer: entitiesDetailsReducer({}),
 };
 
-const middleware = (getDefaultMiddleware) =>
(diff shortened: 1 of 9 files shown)


src/Store/index.js
```diff
@@ -52,22 +49,18 @@ const middleware = (getDefaultMiddleware) =>
     notificationsMiddleware({
       errorTitleKey: ['message'],
       errorDescriptionKey: ['response.data.detail'],
-    }),
-    production && logger
```
bastilian, line 56: That the store is memoized is fine and the `useMemo` can and maybe should stay, but this here seems to be the only line that makes a difference and resolves the error. 

However, I think instead of removing this we should rather be fixing it and enable redux logging properly.

Other apps pass in the the logger depending what entry they are coming from, like here: https://github.com/RedHatInsights/compliance-frontend/blob/master/src/DevAppEntry.js#L5

We can take a similar approach here and avoid using the `env/production` consts here. 

Another thing that compliance does, that may be relevant here is it filters out undefined middlewares passed in. This may help here as well to prevent the issues we were seeing.

mkholjuraev: I do not think that memoization of the store is good practice. The useMemo is not designed for it, I think. Rather, it is [recommended](https://redux-toolkit.js.org/tutorials/quick-start#provide-the-redux-store-to-react) way to assign the initialized store to some constant variable. If the store is a constant var outside of react, it will not be reinitilised on every re-render. This is what Karel wanted to fix with useMemo

bastilian: Still, we should rather fix the passing of the logger, than just removing it, because it is useful in dev.

mkholjuraev: Oh, yes. I will bring it back.

bastilian: I think we'll also need to properly init the logger. See: https://www.npmjs.com/package/redux-logger#usage

src/Store/index.js
```diff
@@ -28,6 +25,21 @@ const reducer = {
...
+  Recs.middleware,
+  Systems.middleware,
+  Topics.middleware,
+  Acks.middleware,
+  SystemVariety.middleware,
+  notificationsMiddleware({
+    errorTitleKey: ['message'],
+    errorDescriptionKey: ['response.data.detail'],
+  }),
+  ...(process.env.NODE_ENV !== 'production' ? [logger] : []),
```
bastilian, line 40: We should use the Entry/DevEntry method. Using the `NODE_ENV` here is too risky.

bastilian: Also, I don't think `process.env` is available in the browser. :)

mkholjuraev: Initially, I tried to go this way. But, I remember entry-dev file is depricated and Compliance [uses](https://github.com/RedHatInsights/compliance-frontend/blob/4e759a201b89e5007b27c8b0c0a59aabd3aa7b04/src/entry-dev.js#L1) it for differentiating prod and dev envs

mkholjuraev:
@bastilian I have implemented the same logic as the compliance app. I have found out following issues:
1. Without useMemo, redux-toolkit does not initialize the store properly. I have tried multiple ways, to fix this root problem. But, no luck. I am blaming the fact that the store gets reinitialized on every re-render without useMemo. The way compliance [init](https://github.com/RedHatInsights/compliance-frontend/blob/master/src/AppEntry.js#L24) the store does not work due to this reason.
2. process.env.NODE_ENV is not available in build time inside fec.config.js in both Compliance and Advisor apps. Effectively, logger middle is not being added in dev env.

Let's get this PR merged if this fixes the stage issues. We need to pass proper NODE_ENV by tweaking the fec related files in the fec-utils repo.

bastilian (dismissed):
LGTM! Thank you @mkholjuraev

bastilian (approved):

mkholjuraev:
cypresssss   🗡️

mkholjuraev:
Finally, cypress tests pass 😎

PR #1275, opened 2024-06-21
fix(RHINENG-10409): Limit justification to backend max length

mhuth:
Recently we had a customer entering a value into the justification field that was longer than the max length of the field, and it had us scratching our heads for a while trying to figure out what the problem was.  We have increased the size of the field in the backend in https://gitlab.cee.redhat.com/insights-platform/advisor-backend/-/merge_requests/887 but we also think we should limit the UI text input to the same max length, to make it harder for customers to 'do the wrong thing'.  This is a simplistic solution and can easily be defeated by determined customers, but its a low-hanging fruit solution to the problem, so I'm going with it.


mkholjuraev (approved):
LGTM! As long as UX is happy, I think this should not make any problem, but make the UI consistent with the backend.

PR #1277, opened 2024-06-26
fix(ZeroStateWrapper): RHINENG-10312 dont show ZeroState on data loading

xmicha82:
# Description

Associated Jira ticket: # [RHINENG-10312](https://issues.redhat.com/browse/RHINENG-10312)

Don't show ZeroState while data are still loading - also fixes persisting Global Filters from Dashboard

# How to test the PR

1. Go to Dashboard application
2. Apply global filter
3. Use any link in Advisor card to go to Advisor application
4. The Global filters should be applied

# Checklist:

### DIFF
--- a/src/ZeroStateWrapper.js
+++ b/src/ZeroStateWrapper.js
@@ -70,7 +70,7 @@ export const ZeroStateWrapper = ({ children }) => {
         </Bullseye>
       }
     >
-      {!hasSystems ? (
+      {!hasSystems && conventionalQuerySuccess ? (
         <Suspense
           fallback={
             <Bullseye>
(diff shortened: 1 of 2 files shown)


mkholjuraev (commented):
The fix works like a charm. I do not see the zero state on load and also global filters are kept when navigated from dashboard app. The only thing is that zero-state has new implementation that needs to be adopted in advisor. Thus, while migrating to that new implementation, this fix could be lost. Can you tell me what is the root cause of this issue? is it only zero state or something else? Let's try to find out more about this and evaluate if this fix is something that helps in the long term before the approval

xmicha82:
@mkholjuraev main cause of the issue was that when the data for advisor were loading, the app fell back to ZeroState, which then removed all parameters from the URL. Making sure that the ZeroState is displayed **only** when there is no systems **and** they finished loading fixed the issue and the link from Dashboard properly propagate to the app

mkholjuraev (approved):
LGTM! We need to make sure not to loose this fix while migrating to new implemenation

PR #1282, opened 2024-09-03
Remove deprecated Component.defaultProps

leSamo:
Fix the following error which was printed into console when running Jest tests:

> Warning: MessageState: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.


mkholjuraev:
@leSamo code wise looks good. We need to fix the tests

PR #1285, opened 2024-09-25
refactor(routing): Updates Advisor Routes to use global schema + Remove unused packages

adonispuente:
This is mainly a routing refactor so that advisor uses the same schema as other apps. Along with some unused package clean ups.


bastilian:
Nice refactor! ✨

There seem to be some minor issues, for example, the "Pathways" tab and the "Immutable ..." tab on the details both lead to broken pages. It might be that we need to updated the paths of the links/navigate there to fix these issues.

Another issue is that the "Zero State" flash up, even if the account has connected systems.

Other than that everything appears to work as before.

adonispuente:
@bastilian Thanks for taking a look! I adding in a loading const to avoid that flash, and it seems I had actually removed some routes on accident. I changed the way we mapped through routes so that it could handle arrays, and also added routes that render the same component into the same object to tidy things up. In the original we have 16 routes, this refactor has 16 as well.

PR #1375, opened 2025-02-20
Fix cypress test collection

opacut:
Fix code coverage scripts and collection so that we have a cypress report generated for codecov.

Review steps:
- npm i
- npm run travis:verify
- check that the coverage-cypress/lcov.info is not empty
- check that the coverage-jest/lcov.info is not empty
- check that .nyc_output/out.json is not empty

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,11 +19,8 @@ yarn-debug.log*
 yarn-error.log*
 
 .DS_Store
-*coverage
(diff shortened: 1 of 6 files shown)


johnsonm325 (approved):
Looks good to me. I'll assume that drop in coverage is probably just the code-coverage correcting itself.

PR #1396, opened 2025-04-08
fix(RHINENG-17315): Show disabled recommendations and systems counts

mhuth:
As discovered in https://issues.redhat.com/browse/RHINENG-17315, when filtering disabled rules, the impacting=false URL parameter was being applied unnecessarily and causing the results to be incorrect.  The Advisor API backend has recently changed how it displays disabled rules, removing the need for impacting=false.

Also, Red Hat disabled rules should show how many systems are being impacted, so I removed the N/A string and tooltip and directly show how many systems are affected.  As per the attached screenshots.

# Checklist:

### DIFF
--- a/src/Messages.js
+++ b/src/Messages.js
@@ -35,7 +35,7 @@ export default defineMessages({
   totalRisk: {
     id: 'totalRisk',
     description:
-      'Recommendation table column title, recommendationdetails label',
+      'Recommendation table column title, recommendation details label',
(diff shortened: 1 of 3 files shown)


mhuth:
Before (the system count change):
Screenshot from 2025-04-08 14-41-41


After (the system count change):
Screenshot from 2025-04-08 16-27-27

PR #1399, opened 2025-04-10
fix(RHINENG-17352): Wrap long justification notes

mhuth:
Another finding from https://issues.redhat.com/browse/RHINENG-17352 was that a long justification note would overflow the containing element, as per attached screenshots.  It is unlikely a customer would do this, but still, it was raised and it was a fairly easy fix since I was already working on that part of the code.
 

mhuth:
Before the fix - long justification note overflows the container:
Screenshot from 2025-04-10 10-03-04


After the fix - long notes are wrapped within the containing element:
Screenshot from 2025-04-10 10-15-08

leSamo (approved):

PR #1404, opened 2025-04-16
fix(RHINENG-17091): Remove 'Disable recommendation' kebab for read-only users

mhuth:
Resolves RHINENG-17091.  Users with read-only permissions are able to click on the button to disable a recommendation for a particular system.  However because they are read-only they shouldn't have such permissions.  Indeed they get a 403 error from the API indicating they don't have permission, but the UI should prevent them from even sending the disable recommendation request to the API.  

This PR removes the 'Disable recommendation' kebab from each row when the user has read-only permissions, as per the screenshots.


mhuth:
Before changes, read-only users can click on the Disable Recommendation kebab and start the process of disabling a recommendation via the modal.  The API then gives a 403 error:

Screenshot from 2025-04-17 09-07-09
Screenshot from 2025-04-17 09-07-21


After the changes, the read-only user isn't shown the Disable Recommendation kebab at all:
Screenshot from 2025-04-17 09-06-55

leSamo (approved):
Nice fix, thank you!

PR #1440, opened 2025-05-27
feat(RHINENG-18208): rbac refactor with usecontext

Fewwy:
# Description

https://issues.redhat.com/browse/RHINENG-18208

Example PR for vulnerability
https://github.com/RedHatInsights/vulnerability-ui/pull/2241

1. Removed usePermissions hook from the App, it is necessary for future work on IOP
2. Added context that now serves as a way to pass permissions throughout the app
3. Added hooks that allow us to map permissions and pass them through the context

# How to test the PR
Test each page that was changed for the functionality that was changed with permissions
Change permissions to true/false to test if the component/functionality is disabled.


mhuth:
@Fewwy LGTM but seems like it needs to be rebased.

PR #1475, opened 2025-06-26
feat(RHINENG-17644): add a way to pass satellite modal as props

Fewwy:
# Description

https://issues.redhat.com/browse/RHINENG-17644
After discussion with Jeremy Lenz he suggested we add a way to pass a remediations modal into our tables
I added that, wrote a test to check that we can pass it and updated some tests configurations to make the future works easier


mhuth:
LGTM :+1:   A couple of minor/optional nitpicks.  Thanks for simplifying the testing!

PR #1520, opened 2025-08-02
feat(RHINENG-17644): add remediate modals and remote hybridinventory tabs for iop

Fewwy:
# Description

https://issues.redhat.com/browse/RHINENG-17644
1. Added 2 functions that fetch data for resolutions in Satellite environment
2. Connected them to the Remediate buttons in "host details" and "recommendation details" in the Satellite environment.
3. Added a separate component for IopRecommendation details that replaces OverviewDetails in the RecommendationDetailsWrapped component. That allowed me to "bypass" the HybridInventoryTabs module from inventory, so we don't need to do any tricks with versions for react-router-dom in Inventory. If we try to use HybridInventoryTabs as before -> we have to either upgrade satellite to react-router-dom v6 or do smth else.
5. Fixed an issue where the Host link in the Inventory table in Recommendation details in the Satellite environment was broken. Now, it leads to the host details for that particular host.

# How to test the PR
1) Test the recommendation details on stage to check that nothing is broken.
2) Test modals in the Satellite environment Host Details -> Insights tab and Recommendation Details tab.

### DIFF
--- a/src/SmartComponents/HybridInventoryTabs/ConventionalSystems/RecommendationSystems.js
+++ b/src/SmartComponents/HybridInventoryTabs/ConventionalSystems/RecommendationSystems.js
@@ -37,7 +37,7 @@ const ConventionalSystems = ({
       exportTable="systems"
       showTags={true}
       axios={axios}
-      IopRemediationModal={props.props.IopRemediationModal}
+      IopRemediationModal={props?.props?.IopRemediationModal}
     />
   );
 };


src/SmartComponents/HybridInventoryTabs/ConventionalSystems/RecommendationSystems.js
```diff
@@ -30,6 +36,8 @@ const ConventionalSystems = ({ rule, afterDisableFn, handleModalToggle }) => {
       permsExport={envContext.isExportEnabled}
       exportTable="systems"
       showTags={true}
+      axios={axios}
+      IopRemediationModal={props.props.IopRemediationModal}
```
bastilian, line 40: This causes an error when running with stage. Should it maybe just be `props.IopRemediationModal`?

PR #1534, opened 2025-08-11
fix(RHINENG-19979): Default to publicPath: auto

mhuth:
In the IOP environment, publicPath: auto must be set.  But apparently this is ok to use for the Hosted environment as well.  However when I tried using it in a local dev environment, the RootApp couldn't be loaded, so I set it to '' when using a local environment with the proxy.  

However I'm not sure if this is the best way to make this work.


bastilian (approved):
LGTM! Thank you @mhuth!

PR #1550, opened 2025-08-28
fix(RHINENG-20343): Use last_seen attribute for Last Seen column

mhuth:
Use the Advisor 'last_seen' attribute for populating the Last seen columns, instead of the inventory 'updated' attribute.  'updated' is updated more often than 'last_seen' and gives a false sense for when a machine was 'last_seen' by Advisor, which should be only when they upload an archive.


LightOfHeaven1994 (approved):
LGTM, all changes looks fine!

PR #1573, opened 2025-09-24
fix(SAT-38669): Display rules table skeleton on first fetch only

mhuth:
isLoading vs isFetching:  isLoading is true on first fetch, isFetching is true on subsequent fetches.  So really, we only want to show the skeleton table on the first fetch and leave it displayed as-is on subsequent fetches.  .... right?

After fix, the skeleton table is displayed on the first fetch only:
Screenshot from 2025-09-24 15-04-52


bastilian:
@mhuth Usually, at least on console.rh.com, we also show the loading/skeleton state when paginating or filtering, etc.

mhuth:
> @mhuth Usually, at least on console.rh.com, we also show the loading/skeleton state when paginating or filtering, etc.

@bastilian Ah yes, that's right.  Hmmm, I'll have to find another way to fix this :thinking:

mhuth:
Closing this and replacing it with https://github.com/RedHatInsights/insights-advisor-frontend/pull/1574

PR #1575, opened 2025-09-29
fix(RHINENG-20837): Add OUIA-ID to components, esp for IOP

mhuth:
Adds data-ouia-component-id values for components that I can, esp Table & Card patternfly components.  However for components from RedHatInsights/frontend-components, most don't take an ouiaId prop so I can't pass an ouiaId value to them.


LightOfHeaven1994 (approved):
LGTM, thanks! 👍

PR #1748, opened 2026-02-11
fix(RHINENG-23894): Hacky FE workaround for BE issue RHINENG-24023

mhuth:
Not proud of this PR but at least it allows the SystemsTable to load in the FE.  It removes systems that don't have a last_seen value, which is an indication that the system only exists in Advisor but not in Inventory.  Thus we don't query Inventory for these systems because Inventory would return a 404 error instead of a list of systems.

I would hope this is just a temporary workaround until Advisor BE issue RHINENG-24023 is resolved (could be a while tho)

Before the fix - in prod as insights-qa and sorting on the Last Seen column:
Screenshot from 2026-02-11 17-22-20

After workaround, the table displays but with a smaller number of systems than a full page:
Screenshot from 2026-02-11 17-23-38

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -262,6 +262,13 @@ const SystemsTable = () => {
             await Get(envContext.SYSTEMS_FETCH_URL, {}, options)
           )?.data;
 
+          // Hack to remove systems without a last_seen value from the table,
+          // because it's likely this system was deleted from Inventory, but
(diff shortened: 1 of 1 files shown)


Fewwy (approved):
Not lgtm yet

Fewwy (commented):
We can't merge this yet, especially if it breaks the way how table works

mhuth:
Closing this PR and using this one instead: https://github.com/RedHatInsights/insights-advisor-frontend/pull/1749

PR #1749, opened 2026-02-11
fix(RHINENG-23894): fix the case with last seen null for systems

Fewwy:
# Description

https://issues.redhat.com/browse/RHINENG-23894

Systems with last_seen: null exist in the Advisor API but not in the Inventory
   API. When the frontend attempts to fetch these systems from Inventory, it
  receives 404 errors that break the UI and prevent users from viewing system   
  tables and details.

  Solution                                                                      
   
  Implemented frontend filtering and error handling across three key areas to   
  prevent 404 errors while maintaining accurate pagination counts.
                                                                                
  Changes Made                                                                  
   
  1. Systems List Page (SystemsTable.js)                                        
                  
  What it fixes: Main systems list at /insights/advisor/systems                 
                  
  Implementation:                                                               
  - Filters systems where last_seen !== null before querying Inventory API
  - Adjusts total count by subtracting filtered systems from current page       
  - Wraps Inventory API call in try/catch to gracefully handle any 404s  
  - Skips Inventory API call entirely if all systems are filtered out           
                                                                                
  2. Recommendation Details - Affected Systems (Inventory/helpers.js)           
                                                                                
  What it fixes: Affected systems tables on recommendation detail pages at      
  /insights/advisor/recommendations/:ruleId/                                    
                                                                                
  Implementation: 
  - Applies identical filtering logic to affected systems tables
  - Prevents users from seeing or clicking on non-existent systems              
  - Adjusts counts and handles errors consistently with SystemsTable
                                                                                
  3. System Detail Pages (InventoryDetails.js)                                  
                                                                                
  What it fixes: Direct navigation to system detail pages at                    
  /insights/advisor/recommendations/:ruleId/:systemId/                          
                                                                                
  Implementation: 
  - Checks if system exists in Advisor API and has valid last_seen before
  rendering                                                                     
  - Shows loading skeleton during validation
  - Displays user-friendly empty state if system doesn't exist                  
  - Shows warning notification explaining why system cannot be displayed        
  - Prevents DetailWrapper component from attempting to load non-existent system
   (which would cause 404)

### DIFF
--- a/src/PresentationalComponents/Inventory/InventoryDetails.js
+++ b/src/PresentationalComponents/Inventory/InventoryDetails.js
@@ -6,7 +6,7 @@ import {
   Grid,
   GridItem,
 } from '@patternfly/react-core/dist/esm/layouts/Grid/index';
-import React, { useContext, useEffect } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { connect, useStore } from 'react-redux';
 import { Title } from '@patternfly/react-core/dist/esm/components/Title/Title';
 import Breadcrumbs from '../../PresentationalComponents/Breadcrumbs/Breadcrumbs';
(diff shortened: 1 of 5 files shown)


mhuth (approved):
Works well! 👍  

Minor suggestion though, just wondering if some of the duplicated code in SystemsTable.js and Inventory/helpers.js could be merged into a common function?

PR #1860, opened 2026-04-14
feat(Kessel): reduce number of API calls

LightOfHeaven1994:
This reverts commit e137ac137de115ecccf86d863ae7982670b2a05e.

Reverting our change based on conversation with [@josejulio ](https://github.com/josejulio). Kessel permissions schema will be changed/fixed so all permissions will be non-host centric

**Upd:**
We found out that for recommendations view role we don't need to check all workspaces but in this case we need to check it with `_assigned` suffix

### How to test:

1. Use `comp-user-3-non-admin` that has advisor view permission + host access but only in Wokrpsace 1
2. Navigate to Advisor Topics and Recommendations and see that page is displayed and self bulk API request returns True for recomendations view

## Summary by Sourcery

Simplify Kessel-based permission checking to use a single workspace context and align relations with the updated recommendations view contract.

Enhancements:
- Streamline Kessel permission resolution to query export, disable recommendations, and recommendations view permissions for the current workspace only.
- Update Kessel relations to use the `_assigned` variant for recommendations view access.

Tests:
- Adjust usePermissionCheck tests to reflect the simplified Kessel permission response structure and updated Kessel relations.
- Remove tests and utilities related to multi-workspace Kessel permission handling and error edge cases now that only the default workspace is checked.


Fewwy (approved):
Great fix, thank you

PR #1864, opened 2026-04-15
feat(RHINENG-25233): topics table tabletools version

Fewwy:
https://redhat.atlassian.net/browse/RHINENG-25233

New version of the table that utilize table tools and can be enabled using feature flag

## Summary by Sourcery

Gate the Topics table behind a feature flag and introduce a new TableTools-based implementation while preserving the existing table as a fallback.

New Features:
- Add a new TopicsTable implementation built on StaticTableToolsTable with column components for name, featured status, and affected systems, plus text-based filtering and sorting.
- Enable switching between the original and new Topics table implementations via the advisor-tabletools-migration feature flag using the Unleash client.

Enhancements:
- Extract the legacy Topics table into a separate TopicsTable.original component to simplify toggling between implementations.
- Introduce a Cypress feature flag interceptor and wrap tests in an Unleash FlagProvider to support flag-dependent behavior in UI tests.

Tests:
- Extend TopicsTable Cypress tests to mount under the feature flag provider and verify rendering for both the new TableTools and legacy implementations based on the flag state.


src/PresentationalComponents/TopicsTable/TopicsTable.new.js
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Topics table using bastilian-tabletools with client-side sorting and filtering.
+ * Displays list of topics with name, featured status, and affected systems count.
+ */
+const TopicsTableNew = ({ props }) => {
+  const intl = useIntl();
+  const { data: topics = [], isLoading, isFetching, isError } = props;
+
+  const columns = [
```
bastilian: It might make sense to extract columns and filters and such into a separate files to keep everything neat. Compliance follows that pattern for almost all tables.

If you need "dynamic" columns or filters, you could wrap them in a hook, like `const columns = useColumns(intl)`

 https://github.com/RedHatInsights/compliance-frontend/tree/master/src/PresentationalComponents/RulesTable

Fewwy: Thank you for the suggestion, I will do that :)

Fewwy: @bastilian I updated pr, now I have hooks for columns and I store filters, columns in a separate file for assets

mhuth (approved):
Seems to function the same as the old topics table.

src/PresentationalComponents/TopicsTable/TopicsTableAssets.js
```diff
@@ -0,0 +1,123 @@
+import React, { useMemo } from 'react';
```
bastilian, line 1: This is good, but I would recommend to split this up and have one file/folder for columns, components and filters, and hooks that use them, similar to how tables in compliance are organised.

bastilian (dismissed):
LGTM! Nice work! Thank you @Fewwy!

src/PresentationalComponents/TopicsTable/TopicsTableAssets.js
Fewwy: Ok, I think I will do that in the nest batch/pr. I want to separate that from this PR, so I can merge this and continue developing

PR #1873, opened 2026-04-22
fix(RHINENG-25967): fixed inventory details issue

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-25967

The system existence check in InventoryDetails.js was calling the wrong API endpoint          
(/system/?system_uuid={id}), which doesn't support the system_uuid query parameter. This      
caused the endpoint to return all systems and validate against a random system's data instead
of the requested one, making all valid systems appear as "not available". 

The fix changes the validation to use the correct /system/{id}/reports/ endpoint, which is designed for checking
specific systems and returns a 404 only when the system truly doesn't exist.

## Summary by Sourcery

Validate inventory system existence using the correct Advisor reports endpoint and add coverage around system availability behavior.

Bug Fixes:
- Correct system existence check in InventoryDetails to call the /system/{id}/reports/ endpoint instead of an unsupported query-based systems endpoint.
- Ensure systems are treated as available on non-404 errors while still surfacing proper messaging and breadcrumbs when a system truly does not exist.

Tests:
- Add comprehensive tests for InventoryDetails to cover successful system lookup, 404 handling and notifications, non-404 error behavior, loading states, and correct use of the reports endpoint and inventoryId from route params.

### DIFF
--- a/src/PresentationalComponents/Inventory/InventoryDetails.js
+++ b/src/PresentationalComponents/Inventory/InventoryDetails.js
@@ -57,16 +57,15 @@ const InventoryDetails = ({ entity }) => {
   useEffect(() => {
     const checkSystem = async () => {
       try {
-        const response = await axios.get(
-          `${envContext.BASE_URL}/system/?system_uuid=${inventoryId}`,
+        await axios.get(
+          `${envContext.BASE_URL}/system/${inventoryId}/reports/`,
         );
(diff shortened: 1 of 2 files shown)


mhuth (approved):
Works well, LGTM 👍  Sorry I missed this problem in a previous review.

PR #1896, opened 2026-04-29
fix(kessel): fixing rbacv1 calls

Fewwy:
# Description
Fixed RBAC v1 race condition in App.js by waiting for feature flags before rendering any context, then cleaned up the PR. added 4 new tests for flag loading states

## Summary by Sourcery

Gate environment context selection on feature flag readiness and route SystemDetail and App through either RBAC v1 or Kessel contexts based on loaded flags.

Bug Fixes:
- Prevent RBAC v1 context from being called before feature flags are loaded or when the Kessel feature is enabled, avoiding race conditions in context initialization.

Enhancements:
- Introduce a dedicated SystemDetailContent component that consumes an injected environment context, with thin wrappers for RBAC v1 and Kessel modes.
- Show a loading spinner in SystemDetail while feature flags are still loading, improving user feedback during initialization.

Tests:
- Expand App and SystemDetail tests to cover feature flag loading states, context selection between RBAC v1 and Kessel, and avoidance of RBAC v1 calls when Kessel is enabled.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -11,6 +11,7 @@ import { useHccEnvironmentContext, useFeatureFlag } from './Utilities/Hooks';
 import { useKesselEnvironmentContext } from './Utilities/useKesselEnvironmentContext';
 import { LockIcon } from '@patternfly/react-icons';
 import { AccessCheck } from '@project-kessel/react-kessel-access-check';
+import { Bullseye, Spinner } from '@patternfly/react-core';
 import { KESSEL_API_BASE_URL } from './AppConstants';
 import { useFlagsStatus } from '@unleash/proxy-client-react';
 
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -82,8 +83,13 @@ const AppWithKesselContext = () => {
 };
 
 const AppWithContextProviders = () => {
+  const { flagsReady } = useFlagsStatus();
   const isKesselEnabled = useFeatureFlag('advisor.kessel_enabled');
 
+  if (!flagsReady) {
+    return null;
```
LightOfHeaven1994: nitpick: @Fewwy maybe we could return spinner here `return <Bullseye><Spinner size="xl" /></Bullseye>;`

LightOfHeaven1994 (approved):
Great work! and Nice catch on SystemDetail component! ❤️

PR #1949, opened 2026-05-15
feat(RHINENG-25234): Pathways tab tabletools migration

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-25234
Turn on feature flag

Open recommendations table -> switch to pathways tab from it, test the UI, apply filters switch pagination etc.
Experience of using pathways should stay the same

TLDR of the changes:
1) I added tanstack query to start migrating our api endpoints to it
2) created new Services/Pathways/ folder (apiClient.js, usePathwaysQuery.js), added QueryClientProvider to AppEntry.js, and built three new utilities (useAdvisorTableDefaults.js for common table options) 
3) Removed intl dependency from PathwaysTable by refactoring Columns.js 
4) Enhanced tableSerializers.js to support per-filter serializers for custom serialization logic. 
5) Updated all test files (Cells.test.js, Columns.test.js, PathwaysTable.test.js, PathwaysTable.cy.js) to remove intl mocking and verify new implementation

URL SYNC is not enabled in new pathways table, it is missing from table tools and will be developed in the near future.


src/PresentationalComponents/PathwaysTable/PathwaysTable.new.js
```diff
@@ -0,0 +1,334 @@
+import React, { useCallback, useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { useLocation } from 'react-router-dom';
+import { TableToolsTable, TableStateProvider } from 'bastilian-tabletools';
+import { useFullTableState } from 'bastilian-tabletools/dist/hooks';
```
mhuth: bastilian-tabletools?  That's the library name?

mhuth (dismissed):
Functionality wise it seems ok.  I couldn't find any problems with it.  I'm just wondering if bastilian-tabletools is the correct library name?

src/PresentationalComponents/PathwaysTable/PathwaysTable.new.js
Fewwy: Yes, it is. you can check it package.json, line 108

src/PresentationalComponents/PathwaysTable/PathwaysTable.new.js
```diff
@@ -0,0 +1,334 @@
...
+  }
+  if (reduxFilters.has_incident && reduxFilters.has_incident.length > 0) {
+    activeFilters['has-incident'] = reduxFilters.has_incident;
+  }
+  if (reduxFilters.reboot_required && reduxFilters.reboot_required.length > 0) {
+    activeFilters['reboot-required'] = reduxFilters.reboot_required;
+  }
+
+  return (
+    <TableToolsTable
```
bastilian, line 64: 100% percent sure that a lot in this file/refactor is actually not needed. You are handling the state, but that is the TableToolsTable's role.

src/PresentationalComponents/PathwaysTable/PathwaysTable.new.js
```diff
@@ -0,0 +1,334 @@
...
+  sortSerialiser,
+  filtersSerialiser,
+} from '../../Utilities/tableSerializers';
+
+/**
+ * Inner table component that uses bastilian-tabletools with server-side operations.
+ * Requires TableStateProvider context for state management.
+ *
+ * Integrates TableToolsTable state with Redux filters and URL params:
+ * - Reads URL params on mount and syncs to Redux
```
bastilian: Don't for use the URL params for now, and don't sync with redux.

src/PresentationalComponents/PathwaysTable/PathwaysTable.new.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const tableOptions = useMemo(
+    () => ({
+      ...advisorTableDefaults,
+      sortBy: { index: 4, direction: 'desc' },
+    }),
+    [advisorTableDefaults],
+  );
+
+  if (error) {
```
bastilian: This is not needed. The TableToolsTable will show an error if the API throws an error.

src/PresentationalComponents/PathwaysTable/PathwaysTable.new.js
```diff
@@ -0,0 +1,92 @@
...
+      </Table>
+    );
+  }
+
+  return (
+    <TableToolsTable
+      items={items || []}
+      columns={columns}
+      total={total || 0}
+      loading={loading}
```
bastilian: You'll only need to pass the items function. The total and the loading state are or should be handled by the TableToolsTable itself.

Fewwy: I don't think that works with Workloads/Tags filter (global filter), when I try to apply it with only "items" function, that function reference doesn't change. So table won't show updated/fetched data.

If I keep 
const { result, loading } = queryResult; 

and this
items={result?.data || []}                                               
total={result?.meta?.total || 0}                                         
loading={loading}
it is properly updating the table with new data.

how can I solve this?

Fewwy: I solved it!

src/PresentationalComponents/PathwaysTable/PathwaysTable.original.js
```diff
@@ -0,0 +1,413 @@
+import {
```
bastilian: Just leave the old file in place. There is no need to copy or rename anything in the old code.

src/PresentationalComponents/PathwaysTable/Columns.test.js
```diff
@@ -0,0 +1,181 @@
...
+import { MemoryRouter } from 'react-router-dom';
+import columns, {
+  Name,
+  Category,
+  Systems,
+  Reboot,
+  RecommendationLevel,
+} from './Columns';
+
+jest.mock('./Cells', () => {
```
bastilian: ? Why

src/PresentationalComponents/PathwaysTable/Cells.test.js
```diff
@@ -0,0 +1,182 @@
+import React from 'react';
```
bastilian: ... This is testing mocks. There is no need to these tests at all.

src/PresentationalComponents/TopicsTable/TopicsTable.new.js
```diff
@@ -25,8 +26,8 @@ const TopicsTableNew = ({ props }) => {
   // eslint-disable-next-line react/prop-types
   const { data: topics = [], isLoading, isFetching, isError } = props;
 
-  const columns = useTopicsColumns();
-  const filters = useTopicsFilters();
+  const tableColumns = columns(intl);
```
bastilian: Why not leave out `intl`? We have decided to deprecate it a long while ago.

Fewwy: Removed intl from topics table as well

src/PresentationalComponents/PathwaysTable/Cells.test.js
Fewwy: Removed it

src/PresentationalComponents/PathwaysTable/Columns.test.js
Fewwy: Yep, as we discussed in the other PR it doesn't make sense

src/PresentationalComponents/PathwaysTable/PathwaysTable.original.js
Fewwy: Yep, did that

src/Services/Pathways/usePathwaysQuery.js
```diff
@@ -0,0 +1,37 @@
+import { useDeepCompareCallback, useDeepCompareMemo } from 'use-deep-compare';
+import { useQueryWithUtilities } from 'bastilian-tabletools';
+import { fetchPathways } from './apiClient';
```
bastilian: Why not use the actual javascript api client generated from the API specs?

Fewwy: Migrated to javascript client

mhuth (changes requested):
Adding this to the PR so it doesn't get lost.  With filtering, if you select both true and false for Reboot required / Incidents, it only sends the first one you select to the API, not both.  If both are selected, it should be like none are selected.  That's what the old pathways table did.  

Otherwise, it seems to work ok.

bastilian (dismissed):
Nice job! Code wise it looks good to me! Though, I think there are still opportunities to refactor more, but that can be done later as well! Thank you @Fewwy!

PR #1989, opened 2026-05-29
feat(RHINENG-25238): rules table tabletools migration

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-25238

This PR adds new RulesTable to use bastilian-tabletools with feature flag advisor-tabletools-migration.
What Changed
- New RulesTable to dynamic TableToolsTable with React Query (useQueryWithUtilities)

## Summary by Sourcery

Migrate RulesTable to the TableTools implementation behind a feature flag while retaining the existing table as a fallback.

New Features:
- Add a TableTools-based RulesTable with server-side filtering, sorting, pagination, expandable details, and rule enable/disable actions.
- Gate the new table implementation behind the advisor-tabletools-migration feature flag.

Bug Fixes:
- Improve rule resolution risk handling for missing resolution data and correctly match supported system types.
- Support filter labels in slug form when serializing table state.

Enhancements:
- Define reusable RulesTable columns, cells, filters, and table actions for the migrated implementation.
- Replace broad Cypress request stubs with RulesTable-specific API interception that applies fixture filtering and pagination behavior.

Build:
- Update tabletools and related frontend dependencies and adjust test coverage exclusions for the new table components.

Tests:
- Update RulesTable Cypress coverage for the TableTools implementation, including filtering, sorting, pagination, exports, permissions, and feature-flag setup.
- Remove legacy URL synchronization and individual filter tests that are not supported by the migrated table implementation.

Chores:
- Remove obsolete table-state parameter combination utilities and tests.


src/Services/Recs/useRecsQuery.test.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
```
bastilian: This test doesn't really do anything.

src/PresentationalComponents/RulesTable/RulesTable.original.js
```diff
@@ -0,0 +1,451 @@
+import './_RulesTable.scss';
```
bastilian: Why copy/rename this? just use the old file/name.

src/PresentationalComponents/RulesTable/RulesTable.new.js
```diff
@@ -0,0 +1,160 @@
+import React, { useMemo, useState, useCallback } from 'react';
```
bastilian, line 1: This test mostly verifies the functioning of the TableToolsTable, which is already tested in the package itself. 

What maybe can be tested in this RulesTable is the `handleDisableClick`, but for that it'd be better to extract it into a desecrate hook and test it there.

src/PresentationalComponents/RulesTable/NewRulesTable.cy.js
```diff
@@ -0,0 +1,738 @@
+import React from 'react';
```
bastilian: Same as with the [RulesTable.new.test.js](https://github.com/RedHatInsights/insights-advisor-frontend/pull/1989/changes#diff-b894909aabde4b08a8e9487482ecb1a64d097a6ec84e7aebfdbcad82475386d6) this test mostly verifies the functioning of the TableTools table features. and they are already tested.

src/PresentationalComponents/RulesTable/Cells.test.js
```diff
@@ -0,0 +1,181 @@
+import React from 'react';
```
bastilian: Unless the cell component has an `if` in it, it honestly doesn't deserve a test, because you'd just verify things that should be verified elsewhere.

cypress/fixtures/newrulestablerecommendations.json
```diff
@@ -0,0 +1,2315 @@
+{
```
bastilian: Can the old fixtures not be used?

src/PresentationalComponents/RulesTable/NewRulesTable.cy.js
Fewwy: Yep, removing that

src/PresentationalComponents/RulesTable/RulesTable.original.js
Fewwy: Yep, removing that

src/PresentationalComponents/RulesTable/Cells.test.js
Fewwy: Yep, removing that

cypress/fixtures/newrulestablerecommendations.json
Fewwy: They can, I removed these new fixtures

package.json
```diff
@@ -58,7 +58,12 @@
       "!src/PresentationalComponents/Cards/OverviewDashbarCard/IopOverviewDashbarCard.js",
       "!src/PresentationalComponents/ExecutiveReport/BuildExecReport.js",
       "!src/PresentationalComponents/Export/SystemsPdfBuild.js",
-      "!src/PresentationalComponents/Export/TablePage.js"
+      "!src/PresentationalComponents/Export/TablePage.js",
+      "!src/PresentationalComponents/RulesTable/RulesTable.new.js",
+      "!src/PresentationalComponents/RulesTable/Cells.js",
+      "!src/PresentationalComponents/RulesTable/Columns.js",
+      "!src/PresentationalComponents/RulesTable/Filters.js",
+      "!src/PresentationalComponents/RulesTable/RuleDetailsWrapper.js"
```
bastilian, line 66: SUS.

bastilian: Don't do this. And we should remove the other exclusions of components as well. This is not at all right.

src/PresentationalComponents/RulesTable/Filters.js
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Impact checkbox filter
+ */
+export const impactFilter = {
+  type: 'checkbox',
+  label: capitalize(FC.impact.title),
+  filterAttribute: FC.impact.urlParam,
+  id: FC.impact.urlParam,
+  urlParam: FC.impact.urlParam,
```
bastilian, line 61: Where are these `urlParams` used. and why do we still need the `FC....` things.

package.json
Fewwy: Haven't you told me before that we don't need to test the stuff that we already testing in Tabletools?
This affects codecoverage, so I added this

src/PresentationalComponents/RulesTable/Filters.js
Fewwy: It is used in src/Utilities/tableSerializers.js
FC things are just naming, I can rename it to FILTER_CATEGORIES if that is preferred :)

package.json
bastilian: @Fewwy Sorry, i didn't see the reply till now. Well, if the other tests are correct you shouldn't need this. There should be tests that render the table, with the cells and columns and filters, but you shouldn't test each filter (type) in each table.

If you want we can sit together and I can maybe show what, at least I, feel deserves a test and what not.

Fewwy: @bastilian Not a problem, I already have some work for rules table based on the comments you left on the Pathways table pr, so I want to upload here before we go for a review.

src/PresentationalComponents/RulesTable/RulesTable.new.js
Fewwy: I extracted handleDisableClick into a hook that is now tested

PR #2005, opened 2026-06-07
fix(RHINENG-26571): add missing access check provider to advisor fed module

Fewwy:
https://redhat.atlassian.net/browse/RHINENG-26751

Advisor was missing access check provider in inventory system details
For example: https://stage.foo.redhat.com:1337/insights/inventory/000f994b-291f-4dd6-b220-b1cacb95d4ce?appName=advisor

1. Added missing AccessCheck.Provider wrapper to the SystemDetail module federation export to enable Kessel permission checks in the inventory integration. 
3. Fixed race condition in SystemAdvisor data fetching by extracting fetch logic into a useAdvisorReports custom hook, which properly manages loading/error states and prevents state updates on unmounted components. 
4. Added propTypes to the exported SystemDetail component for proper prop validation by module federation consumers.

### DIFF
--- a/src/Modules/SystemDetail.js
+++ b/src/Modules/SystemDetail.js
@@ -80,16 +80,6 @@ const SystemDetailWithContextProviders = (props) => {
   );
 };
 
-SystemDetailWithContextProviders.propTypes = {
-  customItnl: PropTypes.bool,
-  intlProps: PropTypes.shape({
-    locale: PropTypes.string,
-    messages: PropTypes.objectOf(PropTypes.string),
-  }),
-  store: PropTypes.object,
-  IopRemediationModal: PropTypes.elementType,
-};
-
 const SystemDetail = (props) => {
   return (
     <AccessCheck.Provider
--- a/src/Modules/SystemDetail.js
+++ b/src/Modules/SystemDetail.js
@@ -101,4 +101,14 @@ const SystemDetail = (props) => {
   );
 };
 
+SystemDetail.propTypes = {
+  customItnl: PropTypes.bool,
+  intlProps: PropTypes.shape({
+    locale: PropTypes.string,
+    messages: PropTypes.objectOf(PropTypes.string),
+  }),
+  store: PropTypes.object,
+  IopRemediationModal: PropTypes.elementType,
+};
+
 export default SystemDetail;
--- a/src/Modules/SystemDetail.js
+++ b/src/Modules/SystemDetail.js
@@ -9,6 +9,8 @@ import { useHccEnvironmentContext, useFeatureFlag } from '../Utilities/Hooks';
 import { useKesselEnvironmentContext } from '../Utilities/useKesselEnvironmentContext';
 import { Bullseye, Spinner } from '@patternfly/react-core';
 import { useFlagsStatus } from '@unleash/proxy-client-react';
+import { AccessCheck } from '@project-kessel/react-kessel-access-check';
+import { KESSEL_API_BASE_URL } from '../AppConstants';
 
 const SystemDetailContent = ({
   customItnl,
@@ -59,7 +61,7 @@ const SystemDetailWithKessel = (props) => {
   return <SystemDetailContent envContext={envContext} {...props} />;
 };
 
-const SystemDetail = (props) => {
+const SystemDetailWithContextProviders = (props) => {
   const { flagsReady } = useFlagsStatus();
   const isKesselEnabled = useFeatureFlag('advisor.kessel_enabled');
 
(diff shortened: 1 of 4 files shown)


src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -479,7 +480,7 @@ const BaseSystemAdvisor = ({
   useEffect(() => {
     const dataFetch = async () => {
       try {
-        const reportsFetch = await axios.get(
+        const reportsResponse = await axios.get(
```
bastilian: Why would `axios.get` return `undefined`?

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
```diff
@@ -489,6 +490,15 @@ const BaseSystemAdvisor = ({
           },
         );
 
+        const reportsFetch = Array.isArray(reportsResponse)
+          ? reportsResponse
+          : reportsResponse?.data || [];
+
+        if (!reportsFetch || reportsFetch.length === 0) {
+          setInventoryReportFetchStatus('fulfilled');
+          return;
```
hugohezel: This early return introduces UX regression for systems that have 0 recommendations - it won't show the NoRecommendations empty state.

Image

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
Fewwy: I added a custom hook to fetch stuff and ensure that its not an issue anymore, so this thing is resolved

src/SmartComponents/SystemAdvisor/SystemAdvisor.js
Fewwy: I made a custom hook to manage the states of the table, so this issue should be fixed

src/SmartComponents/SystemAdvisor/useAdvisorReports.js
```diff
@@ -0,0 +1,58 @@
...
+            headers: {
+              credentials: 'include',
+            },
+            signal: abortController.signal,
+          },
+        );
+
+        const reportsFetch = Array.isArray(reportsResponse)
+          ? reportsResponse
+          : reportsResponse?.data || [];
```
bastilian: When would `reportsResponse` be `undefined`?

Fewwy: Array.isArray check is necessary because useAxiosWithPlatformInterceptors unwraps responses and returns the data directly. you can try removing it and you will see that table will get stuck in the error state.
Do you have any suggestions how to make it better? :)

src/Modules/SystemDetail.js
```diff
@@ -78,6 +80,27 @@ const SystemDetail = (props) => {
...
 
+SystemDetailWithContextProviders.propTypes = {
+  customItnl: PropTypes.bool,
+  intlProps: PropTypes.shape({
+    locale: PropTypes.string,
+    messages: PropTypes.objectOf(PropTypes.string),
+  }),
+  store: PropTypes.object,
+  IopRemediationModal: PropTypes.elementType,
+};
```
johnsonm325: There are duplicate propTypes on both wrapper and wrapped component and it could create a maintenance burden - if props change, both need updating. Do both layers need propTypes or is it or is validation at the boundary sufficient?

johnsonm325: This is minor and I wouldn't hold up merging.

src/SmartComponents/SystemAdvisor/useAdvisorReports.js
johnsonm325: @Fewwy I think @bastilian is trying to say that we may only need:

```
        const reportsFetch = Array.isArray(reportsResponse)
          ? reportsResponse
          : reportsResponse.data || [];
```
with no `?`. Is there a scenario in which `reportsResponse` would be `undefined`? If not, then we don't need that `undefined` check.

johnsonm325 (dismissed):
I left some minor comments, but this looks good.

src/SmartComponents/SystemAdvisor/useAdvisorReports.js
Fewwy: True, I totally missed the point Sebastian was making -___-, I removed optional chaining

src/Modules/SystemDetail.js
Fewwy: Yep, true, I removed them

koalakangaroo:
@johnsonm325 @bastilian @hugohezel  if the code is good to you, could you approve the MR. This MR is critical as the issue impacts all customers in Advisor. Thanks!

bastilian (approved):
LGTM! Thank you @Fewwy for fixing the issue!

PR #2072, opened 2026-07-16
fix(RHINENG-28040): fix freezing bug and update inventory table

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-28040

## Bug Fixes

- `checkRemediationButtonStatus` — rewrote the pathway branch. Original used `=` (assignment) instead of `===` (comparison) when matching `rule_id`, causing every rule to match. Also only checked `resolution_set[0].has_playbook` instead of checking all entries. Now iterates `pathwayRulesList` directly with `.some()` and properly resets disabled state when no match is found.
- `pathwayRulesList` initialized as `{}` but consumed as an array — changed to `[]` to prevent `Array.isArray()` and iteration failures.
- `paginatedRequestHelper` had no error handling — wrapped in try/catch, returns `{ data: [], meta: { count: 0 } }` on failure instead of crashing.
- `getEntities` called `defaultGetEntities` with an empty array when no systems were returned — added early return guard for `systemIds.length === 0`.
- `getEntities` accessed `fetchedSystems.data` and `fetchedSystems.meta.count` without null checks — added `?.` and `|| []` / `|| 0` fallbacks.
- `DownloadPlaybookButton` sent all selected systems for every rule in pathway mode — added `ruleSystemsMap` prop to filter systems per-rule so each rule only targets actually affected systems.

## Enhancements

1. `selectedIds` → `safeSelectedIds` via `useMemo(() => selectedIds || [], [selectedIds])` — eliminates null/undefined checks scattered throughout the component.
2. `selectedIdsRef` — stores `safeSelectedIds` in a ref passed to `getEntities`, preventing stale closure issues in the memoized `fetchSystems`.
3. `fetchSystems` wrapped in `useMemo` — avoids recreating the function on every render, which would trigger unnecessary inventory table refetches.
4. `entities` selector uses `shallowEqual` — prevents re-renders when Redux state changes but entities content is the same.
5. `activeFiltersConfig`, `actionsConfig`, `onLoadHandler`, `tableProps`, `customFilters`, `exportConfig` — all memoized with `useMemo`/`useCallback` to stabilize references and reduce inventory table re-renders.
6. Inline `hideFilters` objects extracted to module-level constants (`CHROMELESS_HIDE_FILTERS`, `STANDARD_HIDE_FILTERS`).
7. `createColumns` null guard — returns `[lastSeenColumn]` when `defaultColumns` is undefined.
8. `rulesCheck` wrapped in try/catch with `console.error`.
9. Duplicate `label` key in export config fixed — was `exportCsv` then `exportJson`, now only `exportJson`.
10. Remediate and Download Playbook buttons wrapped in `Tooltip` — shows explanation on hover when buttons are disabled due to no playbook. Uses `FlexItem` + `div` wrappers for proper layout and disabled-button hover support.
11. Resolutions `useEffect` — added pathway branch that builds IOP resolution data from `pathwayRulesList` × `pathwayReportList`, and clears resolutions on deselect.
12. Chromeless guard — `RemediationButton` returns `null` in chromeless mode when no `IopRemediationModal` is provided.

## Summary by Sourcery

Fix remediation playbook availability checks and stabilize the inventory table behavior in pathway and standard modes.

Bug Fixes:
- Correct remediation button enablement logic in pathway mode to only consider rules with playbooks affecting selected systems.
- Prevent crashes and inconsistent state when fetching inventory systems by adding error handling and null-safe guards in helper functions.
- Avoid sending all selected systems for every rule when downloading playbooks by scoping systems per rule.
- Ensure pathway rule data is treated as an array and handle empty result sets when loading systems.
- Add a null guard for inventory table column creation to avoid failures when default columns are missing.

Enhancements:
- Reduce unnecessary inventory table re-renders by memoizing selectors, callbacks, configs, and table props, and by stabilizing selected ID handling.
- Improve resolutions building for pathway mode so per-host, per-rule remediation data is correctly assembled and cleared when selections change.
- Add tooltips and layout tweaks to remediation and download playbook buttons to explain disabled states and support hover on disabled controls.
- Introduce a chromeless-mode guard to hide the remediation button when no remediation modal is available.
- Add a Bullseye spinner fallback for the async inventory table component for better loading feedback.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -1,9 +1,9 @@
 import './_Inventory.scss';
 
-import React, { useEffect, useState, useCallback, useContext } from 'react';
+import React, { useEffect, useRef, useState, useCallback, useContext, useMemo } from 'react';
 import { TableVariant, sortable, wrappable } from '@patternfly/react-table';
 import { pruneFilters, urlBuilder } from '../Common/Tables';
-import { useDispatch, useSelector, useStore } from 'react-redux';
+import { shallowEqual, useDispatch, useSelector, useStore } from 'react-redux';
 import {
   getEntities,
   allCurrentSystemIds,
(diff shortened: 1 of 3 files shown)


vkrizan (approved):
wfm

PR #2073, opened 2026-07-16
fix(inventory): fix small issues and optimization improvements

Fewwy:
## Bug Fixes

1. **`checkRemediationButtonStatus` rewrite** (`Inventory.js`)
   - Original pathway branch used `.find()` with O(n*m) nested loops over `selectedIds` and `ruleKeys`, and only checked `resolution_set[0].has_playbook` (first entry only).
   - Rewritten to iterate `pathwayRulesList` directly with `for...of`, using `.some()` to check all `resolution_set` entries. Also properly resets disabled state via `setIsRemediationButtonDisabled(!playbookFound)` after the loop instead of only setting it inside.
   - **Why:** Rules with multiple resolutions where the first has no playbook but a later one does were incorrectly disabling the button. The single `setIsRemediationButtonDisabled(!playbookFound)` after the loop guarantees the state is always set, even when no match is found.

2. **Duplicate `label` key in `exportConfig`** (`Inventory.js`)
   - Both chromeless and standard `exportConfig` objects had `label: intl.formatMessage(messages.exportCsv)` followed by `label: intl.formatMessage(messages.exportJson)`. The second silently overwrites the first.
   - Removed the dead `exportCsv` line, keeping only `exportJson`.
   - **Why:** Dead code that hides the intended export label and confuses readers.

3. **`paginatedRequestHelper` error handling** (`helpers.js`)
   - Wrapped the axios call in try/catch. On failure, logs to `console.error` and returns `{ data: [], meta: { count: 0 } }` instead of letting the error propagate.
   - **Why:** An unhandled API failure crashes the entire Inventory table. The fallback allows the component to render an empty state gracefully.

4. **`getEntities` null-safety** (`helpers.js`)
   - Changed `fetchedSystems.data.filter(...)` to `(fetchedSystems?.data || []).filter(...)`.
   - Changed `fetchedSystems.meta.count` to use a `totalCount` variable from `fetchedSystems?.meta?.count || 0`.
   - **Why:** If `paginatedRequestHelper` returns the fallback object or the API returns an unexpected shape, the component no longer crashes on property access.

5. **`createColumns` null guard** (`Inventory.js`)
   - Added `if (!defaultColumns) { return [lastSeenColumn]; }` at the top of `createColumns`.
   - **Why:** When the InventoryTable hasn't loaded columns yet, `defaultColumns` is undefined. Calling `.filter()` on undefined throws. The fallback returns a minimal column set so the table can render.

## Performance

6. **`fetchSystems` memoized with `useMemo`** (`Inventory.js`)
   - Wrapped the `getEntities(...)` factory call in `useMemo` with deps `[pathway, rule, envContext.RULES_FETCH_URL, envContext.SYSTEMS_FETCH_URL, axios]`.
   - **Why:** Without memoization, a new function reference is created every render and passed as `getEntities` prop to InventoryTable, causing it to re-fetch systems on every render cycle.

7. **`safeSelectedIds` + `selectedIdsRef`** (`Inventory.js`)
   - `safeSelectedIds = useMemo(() => selectedIds || [], [selectedIds])` — single null-safe wrapper replacing scattered `selectedIds || []` / `selectedIds?.length` checks.
   - `selectedIdsRef = useRef(safeSelectedIds)` with `.current` updated each render — passed to `getEntities` instead of the raw array.
   - **Why:** `selectedIds` from `useBulkSelect` can be undefined. The ref prevents stale closures in the memoized `fetchSystems` callback — without it, the callback captures the initial `selectedIds` value and never sees updates.

8. **`getEntities` accepts `selectedIdsRef`** (`helpers.js`)
   - Renamed the `selectedIds` parameter to `selectedIdsRef` and dereferences at the top of the inner function: `const selectedIds = selectedIdsRef.current || []`.
   - **Why:** Paired with the ref in Inventory.js. The factory function is memoized, so it needs a ref to read the current selected IDs at call time rather than capturing a stale array from the closure.

9. **`entities` selector with `shallowEqual`** (`Inventory.js`)
   - Added `shallowEqual` as the second argument to `useSelector(({ entities }) => entities || {}, shallowEqual)`.
   - **Why:** Default strict equality causes re-renders whenever any Redux state changes, even if the `entities` object content is identical. `shallowEqual` compares top-level keys, preventing unnecessary re-renders.

10. **`actionsConfig` memoized with `useMemo`** (`Inventory.js`)
    - Converted `getActionsConfig()` (called inline in JSX every render) to a `useMemo` returning the actions object.
    - Deps: `[safeSelectedIds, isRemediationButtonDisabled, resolutions, pathway, pathwayRulesList, pathwayReportList, rule, rulesPlaybookCount, hasPathwayDetails]`.
    - **Why:** Inline `getActionsConfig()` created new React elements and objects every render, forcing InventoryTable to re-process its toolbar actions unnecessarily.

11. **`activeFiltersConfig` memoized with `useMemo`** (`Inventory.js`)
    - Wrapped in `useMemo(() => ({...}), [filters])`.
    - **Why:** Without memoization, a new config object is created every render, causing filter chip re-renders even when filters haven't changed.

12. **`onLoadHandler` extracted to `useCallback`** (`Inventory.js`)
    - The `onLoad` anonymous function (duplicated in both chromeless and standard branches) is now a single `useCallback` with deps `[filters.offset, filters.limit]`.
    - **Why:** Eliminates duplication between the two render branches and prevents `store.replaceReducer` from being called with a new function reference every render.

13. **Memoized table/filter/export configs** (`Inventory.js`)
    - `chromelessTableProps`, `standardTableProps`, `chromelessCustomFilters`, `standardCustomFilters`, `chromelessExportConfig`, `standardExportConfig` — each wrapped in `useMemo`.
    - **Why:** All were inline object literals in JSX, creating new references every render. InventoryTable receives stable props now, reducing child re-renders.

14. **`hideFilters` and `ASYNC_COMPONENT_FALLBACK` extracted to module-level constants** (`Inventory.js`)
    - `CHROMELESS_HIDE_FILTERS`, `STANDARD_HIDE_FILTERS`, and `ASYNC_COMPONENT_FALLBACK` defined outside the component.
    - **Why:** Static objects that never change don't belong inside the component body. Module-level constants are created once and reused, providing stable references.

15. **`columns={createColumns}` passed directly** (`Inventory.js`)
    - Changed from `columns={(defaultColumns) => createColumns(defaultColumns)}` to `columns={createColumns}`.
    - **Why:** The wrapper arrow function is redundant — it just forwards the argument. Passing the callback directly avoids creating a new function reference every render.

## UX Enhancements

16. **Tooltip on disabled Remediate/Playbook buttons** (`Inventory.js`)
    - Both `RemediationButton` (or `IopRemediationModal.WrappedComponent`) and `DownloadPlaybookButton` are wrapped in `<FlexItem><Tooltip><div>...</div></Tooltip></FlexItem>`.
    - Tooltip shows on hover only when `noPlaybookTooltip` is true: systems are selected, buttons are disabled, and data has loaded (pathway: `hasPathwayDetails`, rule: `rulesPlaybookCount >= 0`).
    - Uses `trigger={noPlaybookTooltip ? 'mouseenter focus' : 'manual'}` to conditionally enable/disable the tooltip.
    - The `<div>` wrapper is required because disabled buttons don't receive hover events.
    - **Why:** Users had no feedback explaining why the buttons were greyed out. The tooltip provides context without disrupting the layout (unlike an inline Alert).

## Pathway Improvements

17. **Resolutions `useEffect` pathway branch** (`Inventory.js`)
    - Upstream only built IOP resolution data for single rules (and skipped when `pathway` was truthy). Added a `pathway` branch that iterates `pathwayRulesList` x `pathwayReportList` to build resolution objects for each affected system.
    - Also clears resolutions when no systems are selected (`setResolutions([])`).
    - **Why:** Without this, the IOP remediation modal in pathway mode receives an empty `iopData` array, making it unable to show resolution options for the selected systems.

## Test Updates

18. **`helpers.test.js` — `selectedIds` wrapped in ref objects**
    - All `getEntities(...)` calls in tests updated to pass `{ current: selectedIds }` or `{ current: [] }` instead of raw arrays.
    - **Why:** `getEntities` now expects a ref object (change #8). Tests must match the new interface.

## Summary by Sourcery

Improve the Inventory table’s resilience, performance, and UX around remediation/playbook actions, especially in pathway mode.

Bug Fixes:
- Prevent crashes and incorrect state when inventory API responses fail or return unexpected data by adding error handling and null-safety in helpers and getEntities.
- Ensure remediation button state is correctly derived for pathways by checking all resolutions and all selected systems, and avoid failures when defaultColumns are not yet loaded.
- Fix export configuration so the JSON export label is correctly surfaced without being shadowed by a duplicate CSV label.

Enhancements:
- Reduce unnecessary re-renders and repeated data fetching by memoizing Inventory callbacks, table/filter/export configs, and using shallowEqual for the entities selector.
- Refactor selection handling to use a memoized, null-safe selectedIds array and a ref passed into getEntities so the memoized fetch function always sees the latest selection.
- Expose full IOP resolution data in pathway mode and clear resolutions when no systems are selected so remediation modals behave consistently.
- Extract static filter-hide configs and async fallback UI into module-level constants for reuse and more stable props.
- Add descriptive tooltips to disabled remediation and download playbook buttons to explain why actions are unavailable instead of silently disabling them.

Tests:
- Update helpers tests to pass selectedIds via ref objects to match the new getEntities API signature.


src/PresentationalComponents/Inventory/Inventory.js
```diff
@@ -105,33 +139,46 @@ const Inventory = ({
...
+  const fetchSystems = useMemo(
+    () =>
+      getEntities(
+        handleRefresh,
+        pathway,
+        setCurPageIds,
+        setTotal,
+        selectedIdsRef,
+        setFullFilters,
+        fullFilters,
```
computercamplove, line 155: do you need `fullFilters` here? 
i see you got warning from test runner, same warning i got from vscode
Image

computercamplove: claude: Bug: fullFilters is a dead parameter in getEntities — helpers.js:146, Inventory.js:155
     
  fullFilters is passed into the factory but never read inside the inner async function. The inner function only calls setFullFilters(allDetails) where allDetails is built from config at call time. The outer fullFilters closure value is silently discarded. This is misleading dead code — a future developer may try to "fix" a stale closure that doesn't exist.

computercamplove:
@Fewwy hey, locally i didn't see any issues, the change didn't break anything

- It would be good to address all warnings from Test runner on `inventory.js`
- please consider to get review from someone with deep frontend code knowledge, i think my review is not enough

PR #2102, opened 2026-07-28
feat(RHINENG-26171): add iop mode

adonispuente:
To test this PR, all you need to do:

Run npm i
export IOP_URL={current IOP instance, DM if you need a url} npm run start:proxy:iop
verify everything works as expected after navigating to iop.foo.redhat.com and logging in
Theres currently an issue where when trying to kill the applicaiton, it doesnt always kill webpack instances. If it hangs up, try
ps aux | grep webpack | grep -v grep
and then kill -9 whatever port is running

Theres a currently a PR in FEC to fix this: https://github.com/RedHatInsights/frontend-components/pull/2397

## Summary by Sourcery

Add support and documentation for running Advisor in IOP proxy mode.

New Features:
- Introduce an npm script to run Advisor via an IOP-enabled frontend development proxy.

Enhancements:
- Update frontend components configuration dependency to a newer version and align proxy configuration for IOP usage.

Build:
- Add IOP-specific dev-proxy script wiring environment variables and custom routes configuration.

Documentation:
- Document how to run Advisor in IOP mode via the frontend-development-proxy README section.

PR #2104, opened 2026-07-28
fix(RHINENG-29294): fix 0 systems in rec details bug

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-29294

On foreman-3.16, the systems table sends filter[system_profile][host_type][nil]=true to theAdvisor API, which filters for "conventional" (non-edge) systems by looking for nullhost_type. In a Satellite/Foreman IOP environment, host_type isn't populated the same way as in HCC, so every system gets filtered out — resulting in 0 systems. The fix (already on 3.18/5.0) simplifies it to filter[system_profile]=true, which skips the host type filtering entirely.

how to test:
run the advisor using iop-dev repository or this PR https://github.com/RedHatInsights/insights-advisor-frontend/pull/2102 (you would have to add changes manually to that PR). Open any advisor recommendation to verify that systems table shows the systems instead of an empty state

## Summary by Sourcery

Bug Fixes:
- Fix systems detail and inventory queries that returned zero systems by removing host_type constraints from system_profile filters.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -53,7 +53,7 @@ const Inventory = ({
     offset: 0,
     sort: '-last_seen',
     name: '',
-    'filter[system_profile][host_type][nil]': true,
+    'filter[system_profile]': true,
   });
   const [fullFilters, setFullFilters] = useState();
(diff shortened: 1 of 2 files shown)


adonispuente (approved):
/lgtm! great find

PR #2111, opened 2026-07-29
fix(RHINENG-29339): replace date format with timestamp component

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-29339

Replaced the DateFormat with Timestamp because it was deprecated, the text and tooltips for timestamps should stay the same and look the same
Removed "babel-plugin-lodash": "3.3.4", because it was causing lint issues and its super old, we don't need it anymore

## Summary by Sourcery

Replace deprecated DateFormat usage with PatternFly Timestamp-based date rendering and introduce a shared relative time helper.

Enhancements:
- Unify relative time display across tables and system advisor views using a new toRelativeTime helper.
- Ensure tooltips continue to show full UTC timestamps while inline text remains visually consistent with previous relative date formatting.
- Update recommendation and pathway details views and host acknowledgement modals to use Timestamp with explicit long date formatting.
- Refactor lodash imports to use direct module paths and remove the babel-plugin-lodash dependency from the build configuration.

Tests:
- Adjust inventory helper tests to assert Timestamp-in-Tooltip rendering instead of DateFormat components.
- Update DetailsRules tests to mock Timestamp usage in place of DateFormat.

### DIFF
--- a/src/Utilities/dateHelpers.js
+++ b/src/Utilities/dateHelpers.js
@@ -1,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Timestamp, Tooltip } from '@patternfly/react-core';
+
+const second = 1000;
+const minute = second * 60;
+const hour = minute * 60;
+const day = hour * 24;
+const month = day * 30;
+const year = day * 365;
+
+const formatTime = (number, unit) =>
+  `${number} ${number > 1 ? `${unit}s` : unit} ago`;
+
+const relativeTimeTable = [
+  {
+    rightBound: Infinity,
+    description: (diff) => formatTime(Math.round(diff / year), 'year'),
+  },
+  {
+    rightBound: year,
+    description: (diff) => formatTime(Math.round(diff / month), 'month'),
+  },
+  {
+    rightBound: month,
+    description: (diff) => formatTime(Math.round(diff / day), 'day'),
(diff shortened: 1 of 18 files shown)


LightOfHeaven1994 (commented):
PR looks mostly fine, I found some issue with the tooltip data - before we were showing `hours+minutes + UTC` but now we show `hours+minutes+seconds + GMT`
Image

src/Utilities/dateHelpers.js
```diff
@@ -0,0 +1,80 @@
...
+  if (isNaN(dateObj.getTime())) {
+    return 'Invalid date';
+  }
+
+  const diff = Date.now() - dateObj.getTime();
+
+  return relativeTimeTable.reduce(
+    (acc, { rightBound, description }) =>
+      rightBound > diff ? description(diff) : acc,
+    dateObj.toUTCString().split(',')[1].slice(0, -7).trim() + ' UTC',
```
LightOfHeaven1994: we can move `dateObj.toUTCString().split(',')[1].slice(0, -7).trim() + ' UTC',` up as helper function and reuse it also for Tooltip on line 62 instead of `{dateObj.toUTCString()}`

LightOfHeaven1994 (approved):
LGTM, thank you! ❤️

PR #2150, opened 2026-08-11
feat(RHINENG-29494): add Ansible workload hook for /ansible/advisor route

computercamplove:
# Description

- When Advisor is accessed under the Ansible bundle (/ansible/advisor), the `workload=ansible` URL parameter and filter are now applied automatically so users see Ansible-relevant recommendations and systems by default.
- Adds `useAnsibleWorkloadDefault` hook that detects the Ansible bundle context, injects the workload query parameter into the URL, and provides `defaultFilters` to preserve the workload selection on filter reset.
- Applies to both the Recommendations (`RulesTable`) and Systems (`SystemsTable`) views.

add Ansible workload hook for /ansible/advisor/* route

As part of the global tag filter decommission, this PR replicates the Ansible workload pre-selection that insights-chrome previously injected via `GLOBAL_FILTER_UPDATE` for /`ansible/*` routes. The implementation is per-app (in advisor), not in chrome - as agreed by the Frontend CoP.

# How to test the PR

Before testing - make sure flag is enabled to hide global tag filter https://insights-stage.unleash.devshift.net/projects/default/features/platform.chrome.hide.global-filter

1. Visit /ansible/advisor/systems and /ansible/advisor/recommendations - verify `?workloads=ansible` in URL and "Ansible Automation Platform" chip visible
2. Verify table only shows Ansible systems
3. Add other filters (name, status), click "Reset filters" - verify ansible workload stays, other filters cleared
4. Click X on the ansible chip - verify filter is removed, all systems/rules shown
5. Refresh page - verify ansible filter is re-applied
6. Visit /insights/advsior - verify no workload filter is applied

# Checklist:

## Summary by Sourcery

Apply a default Ansible workload filter for Advisor routes accessed from the Ansible bundle and wire this default into systems and recommendations tables so their filters, URL parameters, and queries consistently reflect the ansible workload context.

New Features:
- Introduce an Ansible workload default hook that applies a workload=ansible filter when accessing Advisor via the Ansible bundle

Enhancements:
- Propagate default workload filters into systems and recommendations tables to align URL parameters, global filters, and table queries
- Update systems and rules tables to accept configurable default filters and adjust workload filtering behavior to use either local or advisor-provided workload values
- Gate rendering of systems and recommendations pages on readiness of the Ansible workload default hook to avoid flicker or inconsistent initial state

### DIFF
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -2,6 +2,7 @@ import './SystemsTable.scss';
 
 import {
   SYSTEM_FILTER_CATEGORIES as SFC,
+  FILTER_CATEGORIES as FC,
   NO_SYSTEMS_REASONS,
 } from '../../AppConstants';
 import React, { useContext, useEffect, useMemo, useState } from 'react';
@@ -40,8 +41,9 @@ import { conditionalFilterType } from '@redhat-cloud-services/frontend-component
 import { EnvironmentContext } from '../../App';
 import { useAddNotification } from '@redhat-cloud-services/frontend-components-notifications/';
 import { useFeatureFlag } from '../../Utilities/Hooks';
+import PropTypes from 'prop-types';
 
-const SystemsTable = () => {
+const SystemsTable = ({ defaultFilters }) => {
   const intl = useIntl();
   const dispatch = useDispatch();
   const store = useStore();
@@ -118,7 +120,11 @@ const SystemsTable = () => {
     delete localFilters.offset;
     delete localFilters.limit;
 
-    return pruneFilters(localFilters, SFC);
+    const filterCategories = localFilters.workload
+      ? { ...SFC, workload: FC.workload }
+      : SFC;
+
+    return pruneFilters(localFilters, filterCategories);
   };
 
   const activeFiltersConfig = {
@@ -132,6 +138,7 @@ const SystemsTable = () => {
           offset: filters.offset,
           hits: ['all'],
           tags: selectedTags,
+          ...(defaultFilters || {}),
         });
       } else {
         itemsToRemove.map((item) => {
@@ -166,7 +173,7 @@ const SystemsTable = () => {
 
   useEffect(() => {
     let combinedFilters;
-    if (search) {
+    if (search || window.location.search) {
       const paramsObject = paramParser();
       paramsObject.tags = selectedTags;
       paramsObject.sort !== undefined &&
@@ -196,6 +203,7 @@ const SystemsTable = () => {
         limit: 20,
         hits: ['all'],
         tags: selectedTags,
+        ...(defaultFilters || {}),
       };
       setFilters(combinedFilters);
     }
@@ -230,7 +238,9 @@ const SystemsTable = () => {
           operatingSystem: false,
           systemTypeFilter: false,
           systemType: false,
-          ...(isWorkloadFilterEnabled && { workloadFilter: false }),
+          ...((isWorkloadFilterEnabled || defaultFilters?.workload) && {
+            workloadFilter: false,
+          }),
         }}
         initialLoading
         autoRefresh
@@ -284,14 +294,13 @@ const SystemsTable = () => {
             true,
           );
 
-          // Local table Workload filter;
-          // separate from the global Chrome tag/workload filter handled by createOptions above
           const localWorkloadFilter = filters?.workloadFilter;
-          if (isWorkloadFilterEnabled && localWorkloadFilter?.length) {
-            Object.assign(
-              options,
-              workloadArrayQueryBuilder(localWorkloadFilter),
-            );
+          const workloadValues =
+            isWorkloadFilterEnabled && localWorkloadFilter?.length
+              ? localWorkloadFilter
+              : advisorFilters?.workload;
+          if (workloadValues?.length) {
+            Object.assign(options, workloadArrayQueryBuilder(workloadValues));
           }
 
           let fetchedSystems;
@@ -413,6 +422,10 @@ const SystemsTable = () => {
       />
     )
   );
+};
+
+SystemsTable.propTypes = {
+  defaultFilters: PropTypes.object,
 };
 
 export default SystemsTable;
--- a/src/PresentationalComponents/SystemsTable/SystemsTable.js
+++ b/src/PresentationalComponents/SystemsTable/SystemsTable.js
@@ -173,7 +173,7 @@ const SystemsTable = ({ defaultFilters }) => {
 
   useEffect(() => {
     let combinedFilters;
-    if (search || window.location.search) {
+    if (search) {
       const paramsObject = paramParser();
       paramsObject.tags = selectedTags;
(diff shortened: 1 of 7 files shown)


src/PresentationalComponents/SystemsTable/SystemsTable.js
```diff
@@ -166,7 +173,7 @@ const SystemsTable = () => {
 
   useEffect(() => {
     let combinedFilters;
-    if (search) {
+    if (search || window.location.search) {
```
Fewwy: The condition search || window.location.search is too broad — it fires BRANCH A for any query
  param in the URL, not just advisor-managed ones. The fix is to either narrow the check or stop
  bypassing React Router.

  Option A: Narrow the check (smallest change)

  // Before (too broad — any query param triggers URL parsing)
  if (search || window.location.search) {

  // After (only trigger for the param we actually care about)
  if (search || new URLSearchParams(window.location.search).has('workload')) {

  This way ?utm_source=email doesn't trigger BRANCH A, but ?workload=ansible does.

  Option B: Let the hook update Redux instead of the URL (cleaner)

  The root problem is that replaceState changes the URL but React Router / Redux don't know
  about it. Instead of reading window.location.search as a workaround, make the hook set Redux
  state directly:

  // In useAnsibleWorkloadDefault.js — return the initial search string
  export const useAnsibleWorkloadDefault = () => {
    const { getBundle } = useChrome();
    const isAnsibleBundle = getBundle() === 'ansible';

    useLayoutEffect(() => {
      if (!isAnsibleBundle) return;
      const url = new URL(window.location);
      if (!url.searchParams.get('workload')) {
        url.searchParams.set('workload', 'ansible');
        window.history.replaceState(window.history.state, '', url.toString());
      }
    }, [isAnsibleBundle]);

    return {
      isReady: true,
      defaultFilters: isAnsibleBundle ? { workload: ['ansible'] } : undefined,
    };
  };

  // In SystemsTable.js — keep the original condition, BRANCH B handles it
  if (search) {
    // BRANCH A: parse URL (only when Redux search is set)
  } else {
    combinedFilters = {
      limit: 20,
      hits: ['all'],
      tags: selectedTags,
      ...(defaultFilters || {}),  // ← BRANCH B now actually runs for ansible
    };
  }

  This way:
  - /ansible/advisor/systems → search is empty → BRANCH B runs → defaultFilters injects { 
  workload: ['ansible'] }
  - /insights/advisor/systems?utm_source=email → search is empty → BRANCH B runs → clean
  defaults, junk param ignored
  - /insights/advisor/systems?sort=name&limit=10 → search has values → BRANCH A runs → URL
  params parsed normally

  Bonus: switch useEffect to useLayoutEffect

  Eliminates the return null blank frame and lets you drop the isReady gate from both List
  components:

  // Before: two frames (blank → content)
  useEffect(() => { /* set URL */ setIsReady(true); }, []);
  // In List.js:
  if (!isReady) return null;  // blank frame

  // After: one frame (content immediately)
  useLayoutEffect(() => { /* set URL */ }, []);
  // isReady always true, no gate needed in List.js

  Option B + useLayoutEffect is the recommended approach — it's ~10 lines of change, removes the
  window.location.search workaround entirely, and makes defaultFilters actually do what it was
  designed to do.

src/Utilities/hooks/useAnsibleWorkloadDefault.js
```diff
@@ -0,0 +1,39 @@
...
+  const isAnsibleBundle = getBundle() === ANSIBLE_WORKLOAD;
+
+  const url = new URL(window.location);
+  const hasWorkloadParam = url.searchParams.has(WORKLOAD_PARAM);
+
+  const [isReady, setIsReady] = useState(!isAnsibleBundle || hasWorkloadParam);
+
+  const appliedRef = useRef(false);
+
+  useEffect(() => {
```
Fewwy: useEffect runs after the browser paints, so for one frame the user sees nothing because the
  component returns null.

  What happens now

  Frame 1 (paint):
    Hook: isReady = false
    List component: if (!isReady) return null  →  blank screen painted

  Frame 2 (after paint):
    useEffect fires → sets URL → setIsReady(true) → re-render
    List component: if (!isReady is now true) → render page content

  On fast machines that's ~16ms of blank. You can't see it with the naked eye. But it is a real
  Cumulative Layout Shift — the page goes from empty to full content in one jump.

  The fix: useLayoutEffect

  // Before
  import { useEffect, useRef, useState } from 'react';

  export const useAnsibleWorkloadDefault = () => {
    const { getBundle } = useChrome();
    const isAnsibleBundle = getBundle() === 'ansible';
    const [isReady, setIsReady] = useState(!isAnsibleBundle || hasWorkloadParam);
    const appliedRef = useRef(false);

    useEffect(() => {
      // ... set URL, then setIsReady(true)
    }, [isAnsibleBundle]);

    return { isReady, defaultFilters };
  };

  // After
  import { useLayoutEffect, useRef } from 'react';

  export const useAnsibleWorkloadDefault = () => {
    const { getBundle } = useChrome();
    const isAnsibleBundle = getBundle() === 'ansible';
    const appliedRef = useRef(false);

    useLayoutEffect(() => {
      if (!isAnsibleBundle || appliedRef.current) return;
      appliedRef.current = true;
      const url = new URL(window.location);
      if (!url.searchParams.get('workload')) {
        url.searchParams.set('workload', 'ansible');
        window.history.replaceState(window.history.state, '', url.toString());
      }
    }, [isAnsibleBundle]);

    return {
      isReady: true,  // always true — URL is set before paint
      defaultFilters: isAnsibleBundle ? { workload: ['ansible'] } : undefined,
    };
  };

  What changes

  Frame 1 (before paint):
    useLayoutEffect fires → sets URL → done

  Frame 1 (paint):
    isReady is always true → render page content immediately

  No blank frame. No useState for isReady. No if (!isReady) return null in the List components.

  Why useLayoutEffect is safe here

  useLayoutEffect blocks paint, so you don't want expensive work in it. But this hook does two
  cheap operations:
  - new URL(window.location) — microseconds
  - window.history.replaceState(...) — microseconds

  No API calls, no DOM measurement, no heavy computation. Blocking paint for microseconds is
  invisible.

  What you can remove from the List components

  // Before (Systems/List.js and Recs/List.js)
  const { isReady, defaultFilters } = useAnsibleWorkloadDefault();

  if (!isReady) {
    return null;     // ← delete these two lines
  }

  // After
  const { defaultFilters } = useAnsibleWorkloadDefault();
  // isReady no longer needed — just render normally

  Three lines deleted from each List component, isReady state eliminated from the hook. Simpler
  code, no flash.

Fewwy (changes requested):
I did a review and run claude with an attempt to find some edge-cases or issues for the code you introduced. Leaving 2 big comments with an examples of my concerns and how to fix them

src/Utilities/hooks/useAnsibleWorkloadDefault.js
computercamplove: @Fewwy changes are applied, thank you for the suggetions!

Fewwy (approved):
LGTM

PR #2152, opened 2026-08-12
fix(RHINENG-29782): null safety and error handling in helpers js

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-29782

Wrapped paginatedRequestHelper in a try/catch so that network failures and server errors return an empty fallback instead of crashing the Inventory table. Also added optional chaining and fallbacks on fetchedSystems.data and fetchedSystems.meta.count in getEntities to handle cases where the API returns an unexpected shape like null or missing fields. Without these guards, inventory table is prone to crashing.

## Summary by Sourcery

Harden inventory helper data loading against request failures and unexpected API responses.

Bug Fixes:
- Prevent inventory data loading from crashing when paginated requests fail or return null, missing, or malformed response fields.
- Return empty inventory results and zero totals when systems data cannot be fetched.

Tests:
- Add coverage for network and server request failures, null or missing API response fields, and empty response objects.

### DIFF
--- a/src/PresentationalComponents/Inventory/helpers.js
+++ b/src/PresentationalComponents/Inventory/helpers.js
@@ -60,20 +60,25 @@ export const paginatedRequestHelper = async ({
     Object.assign(options, workloadArrayQueryBuilder(localWorkloadFilter));
   }
 
-  return pathway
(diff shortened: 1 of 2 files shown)


bastilian (approved):
LGTM! Thank you @Fewwy!

PR #2153, opened 2026-08-12
fix(RHINENG-29784): inventory table createColumns null guard

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-29784

Added a null guard to createColumns so it returns a minimal [lastSeenColumn] fallback when defaultColumns is undefined, which happens before the InventoryTable finishes loading its columns. Without this, calling .filter() on undefined crashes the page on initial render.

## Summary by Sourcery

Guard inventory column creation during loading so the table renders safely before its default columns are available.

Bug Fixes:
- Prevent the inventory page from crashing during initial column loading when column definitions are unavailable.
- Provide a last-seen column fallback until the inventory table's columns are ready.

Tests:
- Add coverage for undefined, null, and populated inventory column definitions.

### DIFF
--- a/src/PresentationalComponents/Inventory/Inventory.js
+++ b/src/PresentationalComponents/Inventory/Inventory.js
@@ -322,6 +322,9 @@ const Inventory = ({
 
   const createColumns = useCallback(
     (defaultColumns) => {
+      if (!defaultColumns) {
+        return [lastSeenColumn];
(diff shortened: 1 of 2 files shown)


bastilian (approved):
LGTM! Thank you for fixing this @Fewwy!

PR #2155, opened 2026-08-12
fix(RHINENG-26564): fix topics table name filter and add tests

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-26564

Fix edge cases for name filter in topics table and add tests to verify the fix

## Summary by Sourcery

Correct topic name filtering for regex-sensitive search terms and verify the behavior with expanded table tests.

Bug Fixes:
- Fix topic name filtering so special characters and partial names are handled correctly without breaking searches.

Tests:
- Add coverage for topic searches containing parentheses, dots, brackets, plus signs, asterisks, and pipe characters.

### DIFF
--- a/cypress/fixtures/topics.json
+++ b/cypress/fixtures/topics.json
@@ -34,5 +34,14 @@
         "featured": false,
         "enabled": true,
         "impacted_systems_count": 694
+    },
+    {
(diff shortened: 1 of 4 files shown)


bastilian (approved):
LGTM! Filter works as expected! Thank you @Fewwy!

PR #2164, opened 2026-08-14
fix(RHINENG-29357): systems table os filter fix and add docs

Fewwy:
# Description

https://redhat.atlassian.net/browse/RHINENG-29357

Fixes several bugs in the Systems table filters:

- OS filter rhel_version array handling — rhel_version was being passed as an array to Qs.stringify which serialized it as repeated query params instead of the comma-separated format the API expects. Fixed by joining the array into a comma-separated string in createOptions. Also hardened pruneFilters with optional chaining so it doesn't crash when a filter category has no predefined values.

- "Total Risk" filter chip showed inconsistent capitalization because the title was already capitalized before capitalize() ran on it. Changed the source title to lowercase so capitalize() produces the correct result.

- Workload filter chip was missing from the Systems table because workload wasn't included in SYSTEM_FILTER_CATEGORIES. Added it with a conditional so it only appears when the feature flag is enabled.

- URL pollution — handleRefresh was spreading all API options into the URL, including internal keys that don't belong there. Replaced with a whitelist of valid URL params.

Also adds documentation for the OS filter integration flow between advisor and inventory, and new tests for pruneFilters and createOptions.

How to test this

1. Open https://stage.foo.redhat.com:1337/insights/advisor/systems
2. Verify:
  - Select OS filter versions — rhel_version appears in the URL as comma-separated values
  - Clicking a major OS version (e.g. "RHEL 8") selects/deselects all its minor versions
  - "Total risk" chip displays with correct capitalization
  - Workload filter chip appears when feature flag is enabled
  - URL only contains valid filter params (no limit, offset, filter[system_profile], etc.)

## Summary by Sourcery

Fix Systems table filter handling and document the operating-system filter integration.

New Features:
- Add workload filter support to the Systems table when enabled by its feature flag.

Bug Fixes:
- Correctly serialize operating-system version filters as comma-separated API and URL values.
- Prevent filter chip generation from failing for dynamically defined categories without predefined values.
- Correct Total risk chip capitalization and prevent internal API options from polluting the URL.

Documentation:
- Document the Advisor and Inventory operating-system filter integration, including state, URL, API, and chip flows.

Tests:
- Add coverage for operating-system filter serialization, filter precedence, empty values, and resilient filter chip generation.

### DIFF
--- a/docs/os-filter-integration.md
+++ b/docs/os-filter-integration.md
@@ -1,0 +1,414 @@
+# Operating System Filter: Advisor + Inventory Integration
+
+## Overview
+
+The OS filter on the Advisor Systems table (`/insights/advisor/systems`) is a cross-application integration between `insights-advisor-frontend` and `insights-inventory-frontend`. The filter UI is owned by Inventory (loaded via Module Federation / Scalprum), while filter persistence (URL sync, API calls) is handled by Advisor.
+
+This document explains the full data flow for selection, deselection, URL synchronization, and API communication.
+
+---
+
+## Architecture
+
+```
+┌─────────────────────────────────────┐
+│        insights-advisor-frontend     │
+│                                      │
+│  SystemsTable.js                     │
+│    ├── Redux: filters.sysState       │
+│    ├── customFilters.advisorFilters   │
+│    ├── getEntities() → API call      │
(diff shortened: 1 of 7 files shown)


bastilian (approved):
LGTM! Thank you @Fewwy!

